PR #23, opened 2018-10-18
IAM_roles

ttrzcinski:
Added a method calling AWS in order to obtain roles following:

// TODO for a reasonable application you need to retrieve the roles of the user from a central IAM system


hohwille:
was replaced with #24.

PR #24, opened 2018-10-18
Added obtaining of IAM roles.

TomaszTen:
Fill up for IAM roles obtaining following previous TODO.

### DIFF
--- a/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
+++ b/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
@@ -5,6 +5,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
+import java.util.List;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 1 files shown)


templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
```diff
@@ -79,14 +80,64 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
...
+      }
+    } catch (Exception exc_1) {
+      if (showErrorOutput) {
+        exc_1.printStackTrace();
+      }
+    } finally {
+      if (reader != null) {
+        try {
+          reader.close();
+        } catch (Exception exc_2) {
```
hohwille: you could use try-with-resource syntax to reduce your code and make it simpler and more compact.
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
```diff
@@ -79,14 +80,64 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
...
+    Collection<String> roles = new ArrayList<String>();
+
+    final String baseCommand = "aws iam list-groups-for-user --user-name";
+    String command = String.format("%s %s", baseCommand, userName);
+
+    BufferedReader reader = null;
+    Process process;
+    try {
+      process = Runtime.getRuntime().exec(command);
+      process.waitFor();
```
hohwille: so you have changed the app to connect to AWS. Nice feature.
But should we have that by default? What if project does not want to use AWS? What if "aws" command is not available?
Do not get me wrong: I love the idea, but this could be an option or maybe we need more like a documentation having that as  a code-snippet, or we could add some minimum module code for this that can be opted in.
See also:
https://github.com/oasp/oasp4j/issues/633

hohwille:
Seems outdated and no further feedback. Hence, I am closing. Limiting our app template to AWS seems not to be an option. Feel free to reopen/rework...

PR #30, opened 2018-10-29
Added MasterCto for reuse

hohwille:
test to see what is conflicting here, Git seems to be broken. Simple and single commit can not be created as PR resp. is not mergable.


hohwille:
Closing as entirely broken/messed by git merging. Replaced with PR #32.

Issue #33, opened 2018-11-01
Property attribute is  required in PageableJsonDeserializer while deserialize it


vapadwal:
This issue is reference to PR https://github.com/devonfw/devon4j/pull/29

while testing mythaistar it is found that "property" attribute is undefined in Pageable/Sort when loading the menu first time. So I think its best to defined "property" as optional.

As in the PR https://github.com/devonfw/devon4j/pull/29 it states that "property" is a required value.

PR #34, opened 2018-11-01
Fixed for Issue Property attribute is required in PageableJsonDeserializer while deserialize it #33

vapadwal:
This PR is fix for issue https://github.com/devonfw/devon4j/issues/33


vapadwal:
Yes Jorg you were correct, mythaistar angular was passing sort attribute to server as a filter, I have fixed that in angular part, so I am closing this PR

PR #35, opened 2018-11-06
externalize PGP key

hohwille:
simply add profile to `.m2/settings.xml` with property `gpg.keyname` to configure your key

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -485,9 +485,6 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-gpg-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


hohwille:
As we missed 3.0.0 due to lack of review, we should at least have this in 3.1.0. Please give review feedback. It is a very simple change.

hohwille:
If no-one is going to review this, I will be forced to do it myself. However, this is discouraged and I would like to keep our QS process.
Please review!

hohwille:
As nobody was able to review the 3 lines of XML in 4 months I merge this without a review.

Issue #7, opened 2018-09-20
use logstash-logback-encoder (JSON logging)


hohwille:
For logging eiPaaS is using ELK and today I learned that they have standardized the log-format and require apps to follow this. This technically means they use this standard OSS encoder for logback which produces JSON log lines instead of regular text patterns:
https://github.com/logstash/logstash-logback-encoder
@amarinso stated:

> I would vote for having a single log file, json formatted and tools (or tips or techniques ) to split it if required or filter/prettyprint. We should be pushing for systems where log aggregators are the new normal. Let's make modern techniques, principles and patterns our default blueprint.

@maihacke stated:

> I totally agree. Having a central log-collection is normal today and e.g. a security requirement in German public sector projects. So we should go with the JSON-Format per default.
> I do so and have a special log-config for development where it is deactivated.

amarinso:
We should test if the JSON output formatting adds any noticeable performance penalty for an app, so in that case we can even go for async appenders. I read out of curiosity that in logstash they are using LMAX disruptor for enqueing the messages... nice!

hohwille:
merged PR #36 hence done.

hohwille:
> We should test if the JSON output formatting adds any noticeable performance penalty for an app, so in that case we can even go for async appenders. I read out of curiosity that in logstash they are using LMAX disruptor for enqueing the messages... nice!

I do not see a real issue with the log messages we are producing. In production regular mode will most probably even to raise the loglevel to INFO.
However, your knowledge/observation about async appenders sounds very interesting. Please feel invited to add a link or some example config to the documentation:
https://github.com/devonfw-wiki/devon4j/wiki/guide-logging

PR #36, opened 2018-11-06
#7: use json logging for non-dev usage by default

hohwille:
Here is my PR for #7

### DIFF
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -255,13 +255,21 @@
         <artifactId>javax.el</artifactId>
         <version>2.2.6</version>
       </dependency>
-
+      
+      <!-- JSON logging for ELK, etc. -->
+      <dependency>
+        <groupId>net.logstash.logback</groupId>
+        <artifactId>logstash-logback-encoder</artifactId>
+        <version>5.2</version>
+      </dependency>
+    
       <!-- Optional logging dependency -->
       <dependency>
         <groupId>org.owasp</groupId>
         <artifactId>security-logging-logback</artifactId>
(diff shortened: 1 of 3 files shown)


hohwille:
BTW: I have meanwhile done some more intensive testing. CorrelationId is properly working:
```json
{"timestamp":"2018-11-06T18:36:37.572+00:00","@version":"1","message":"Interceptors contributed by service: []","logger_name":"org.apache.cxf.interceptor.OutgoingChainInterceptor","thread_name":"http-nio-8081-exec-6","level":"DEBUG","level_value":10000,"correlationId":"12e7842d-05f2-4253-8821-fa40c28a1d9e","appname":"basic"}
```
One issue that I found is that correlationId is not present for logs for performance logging:
```json
{"timestamp":"2018-11-06T18:36:37.638+00:00","@version":"1","message":"http://localhost:8081/services/rest/security/v1/currentuser;6678;200;;","logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter","thread_name":"http-nio-8081-exec-6","level":"INFO","level_value":20000,"appname":"basic"}
```
This is due to the fact that the `DiagnosticContextFilter` is in the wrong filter order compared to the `PerformanceLogFilter` and hence removes the `correlationId` from the MDC before the performance data is logged.
Second it would be awesome to find a way that via SLF4J we could log custom fields such as status-code or respons-time for the performance logging would end up as separate fields in the JSON. This would be very powerful for kibana dashboards, etc. At the moment we are producing CSV in the message was is not so great. I will file an issue for this but we will not be able to address this before the release.

Issue #37, opened 2018-11-08
devon4j api pom should use spring dependencies


jdiazgon:
As the new `devon4j-api` is used to contain api related files like interfaces, I was trying to generate with CobiGen those templates there, so that we are compliant with the new architecture.

One of the templates that I want to generate on the `devon4j-api` project is `SearchCriteriaTo`. This class extends from `AbstractSearchCriteriaTo` which imports `org.springframework.data.domain.Pageable;`.

The problem is that, by default, the pom of `devon4j-api` does not define the spring related dependencies like:
```xml
    <dependency>
      <groupId>com.devonfw.java.starters</groupId>
      <artifactId>devon4j-starter-spring-data-jpa</artifactId>
    </dependency>
```

Therefore we get compilation errors as Pageable is not found.

I would like to have that dependency on the pom by default.

hohwille:
@jdiazgon thanks for this excellent finding.
As the `api` should be minimal from dependencies point of view do not put a dependency on the starter but use `spring-data-commons` instead. Thanks.

hohwille:
Please fix here:
https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/api/pom.xml#L15
If you could provide a PR would be awesome.

hohwille:
Merged PR #40, hence done.

PR #40, opened 2018-11-09
Fix spring dependencies in api (issue #37)

cmicoege:
Adresses issue #37.
Changed the pom file in: https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/api/pom.xml#L15

Adding the dependency:
<dependency>
    <groupId>org.springframework.data</groupId>
    <artifactId>spring-data-commons</artifactId>
    <version>2.1.2.RELEASE</version>
</dependency>

### DIFF
--- a/templates/server/src/main/resources/archetype-resources/api/pom.xml
+++ b/templates/server/src/main/resources/archetype-resources/api/pom.xml
@@ -14,6 +14,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.springframework.data</groupId>
+      <artifactId>spring-data-commons</artifactId>
+      <version>2.1.2.RELEASE</version>
+    </dependency>
+    <dependency>
       <groupId>com.devonfw.java.modules</groupId>
       <artifactId>devon4j-rest</artifactId>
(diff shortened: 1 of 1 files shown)


templates/server/src/main/resources/archetype-resources/api/pom.xml
```diff
@@ -13,6 +13,11 @@
   <description>API of the server for the ${rootArtifactId} application (containing datatypes, transfer-objects, and service interfaces).</description>
 
   <dependencies>
+    <dependency>
+      <groupId>org.springframework.data</groupId>
+      <artifactId>spring-data-commons</artifactId>
+      <version>2.1.2.RELEASE</version>
```
hohwille: Version already comes from BOM and should not be specified redundantly here as we might miss that on upgrades and have additional maintenance effort.

hohwille:
I merged this as planned.
Seems that when squashing the merge github is unable to detect that the PR has been merged.

hohwille:
@cmicoege thank you for the fix.

PR #42, opened 2018-11-13
Migrated oasp4j wiki references to new devon4j wiki

AbhayChandel:
Apart from the main task 

- Did some remaining namespace migration.
- Code cleanup


AbhayChandel:
I thought of the same thing when migrating the links. But as you said this fine for now and can be taken care of in coming weeks.

Issue #45, opened 2018-12-04
NPE in RestServiceExceptionMapper


hohwille:
If for some reason the service does not provide a UUID in failure JSON, the RestServiceExceptionMapper will do `UUID.fromString(null)` what will end up in a NPE:
https://github.com/devonfw/devon4j/blob/5a42abbc8a6afa8488d8e249ed445702dc991c15/modules/cxf-client-rest/src/main/java/com/devonfw/module/cxf/common/impl/client/rest/RestServiceExceptionMapper.java#L67

The code should be changed to handle the case that no UUID is provided.

hohwille:
PR has been merged. Hence done.

PR #48, opened 2019-01-10
Fixed NPE for uuid instance

AbhayChandel:
PR for https://github.com/devonfw/devon4j/issues/45


hohwille:
@AbhayChandel thanks for the fix 👍

Issue #38, opened 2018-11-08
Change order of PerformanceLogFilter and DiagnosticContextFilter


hohwille:
as stated in #36 the performance logging does not include the `correlationId`:
```json
{"timestamp":"2018-11-06T18:36:37.638+00:00","@version":"1","message":"http://localhost:8081/services/rest/security/v1/currentuser;6678;200;;","logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter","thread_name":"http-nio-8081-exec-6","level":"INFO","level_value":20000,"appname":"basic"}
```
This is due to the fact that the `DiagnosticContextFilter` is in the wrong filter order compared to the `PerformanceLogFilter` and hence removes the `correlationId` from the MDC before the performance data is logged.

Acceptance criteria:
correlationId is included in the performance logging.

AbhayChandel:
Fixed as part of PR https://github.com/devonfw/devon4j/pull/50

PR #50, opened 2019-01-16
Ordered Filters to log Correlation Id in performance logging

AbhayChandel:
Issue: https://github.com/devonfw/devon4j/issues/38

About this change:

1. _Correlation Id_ will be logged conditionally(look point 3) as part of performance logging.
2. In Spring Boot, default order for filters is `0`. So in order to ensure the availability of _Correlation Id_ in `PerformanceLogFilter`, order for `PerformanceLogFilter` has been set to `2` and that for `DiagnosticContextFilter` to `1`.
3. Since `DiagnosticContextFilter` is mapped to _URL_ `/services/*`, the _Correlation Id_ will be logged as part of performance logging only for the aforementioned _URL_ and not for all _URL_s. 

**Update 12-feb-2019**
After the update following use cases were added
1. Now _logback.xml_ configuration will be available under src/main/resources for _Core_ module in _application-template-server_.
2.  _logback.xml_ will not be part of Maven generated JAR for Core module.


modules/logging/src/main/java/com/devonfw/module/logging/common/impl/PerformanceLogFilter.java
```diff
@@ -89,8 +95,11 @@ private void logPerformance(ServletResponse response, long startTime, String url
       errorClass = error.getClass().getName();
       errorMessage = error.getMessage();
     }
-    String message =
-        createMessage(url, Long.toString(duration), Integer.toString(statusCode), errorClass, errorMessage);
+    String correlationId = this.diagnosticContextFacade.getCorrelationId();
+    String message = correlationId != null
+        ? createMessage(correlationId, url, Long.toString(duration), Integer.toString(statusCode), errorClass,
+            errorMessage)
+        : createMessage(url, Long.toString(duration), Integer.toString(statusCode), errorClass, errorMessage);
```
hohwille: This is a missunderstanding. The magic trick of the correlation-id and the MDC is that the logger add this automatically to the log without the logging code line to care about this.
So if it is still present in the MDC it would be logged twice now:
https://github.com/devonfw/devon4j/blob/develop/modules/logging/src/main/java/com/devonfw/module/logging/common/impl/DiagnosticContextFacadeImpl.java#L25

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/service/impl/config/WebConfig.java
```diff
@@ -59,6 +60,7 @@ public FilterRegistrationBean diagnosticContextFilter() {
     this.beanFactory.autowireBean(diagnosticContextFilter);
     registration.setFilter(diagnosticContextFilter);
     registration.addUrlPatterns(ServiceConstants.URL_PATH_SERVICES + "/*");
+    registration.setOrder(1);
```
hohwille: these two should be the actual fix. A nicer option is to use "add filter before/after" mechanism of spring.
How did you test this?

hohwille: As discussed already this add filter before/after is only available for spring-security filter chain. We do not want to couple with spring-security here so this fix should be fine.

modules/logging/src/main/java/com/devonfw/module/logging/common/impl/PerformanceLogFilter.java
hohwille: Can you confirm? Could you undo this change?

AbhayChandel: Yes. I will take this up tomorrow.

AbhayChandel: I have updated this PR with latest code changes as well as my opening comment to include new use cases.

PR #51, opened 2019-01-17
WIP: addition of module for jwt and kerberos

ssarmokadam:
This PR includes module for jwt and kerberos for authentication of user. I was able to test JWT module successfully. In kerberos I have created dummy userdetails which try to login in kerberos  using JAAS. Here I was facing some issues. So some changes might be needed in kerberos module.


vapadwal:
closing this as  create issue for the same -https://github.com/devonfw/devon4j/issues/262

PR #61, opened 2019-02-19
#60 fix - mock web application context rather then remove it on component tests

### LINKED ISSUE #60: ApplicationComponentTest interferes with DiagnosticContextFilter
Inheriting ApplicationComponentTest, which sets `webEnvironment = NONE`, tests fail as of the WebConfig shipped by the archetype creates a diagnosticContextFilter, which needs to get the WebApplicationContext injected. So every test fails as there is no WebApplicationContext.

A solution would be to remove webEnvironment from the SpringBooTest annotation of ApplicationComponentTest, which would leed SpringBoot mock the WebApplicationContext by default.

maybeec:
Resolves #60 as written in the issue itself.

### DIFF
--- a/templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/ApplicationComponentTest.java
+++ b/templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/ApplicationComponentTest.java
@@ -10,7 +10,7 @@ import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
 /**
  * Abstract base class for {@link ComponentTest}s of this application.
  */
-@SpringBootTest(classes = { SpringBootApp.class })
+@SpringBootTest(classes = { SpringBootApp.class }, webEnvironment = WebEnvironment.MOCK)
 public abstract class ApplicationComponentTest extends ComponentTest {
 
   @Override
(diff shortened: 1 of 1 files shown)


templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/ApplicationComponentTest.java
```diff
@@ -10,7 +10,7 @@
 /**
  * Abstract base class for {@link ComponentTest}s of this application.
  */
-@SpringBootTest(classes = { SpringBootApp.class }, webEnvironment = WebEnvironment.NONE)
+@SpringBootTest(classes = { SpringBootApp.class })
```
hohwille: not very explicitly but convention over configuration (default is `WebEnvironment.MOCK`).

hohwille:
I will squash merge this but would still love to get an answer to my above question fur curiostity.

maybeec:
Which above question?

hohwille:
> Which above question?

However, our tests are not failing. Is that simply because we deliver our archetype without any component test in place? I am slightly confused as we have tests for the archetype and I would like to understand what is going wrong here. Did this never work? I have done many projects with devon and written component tests derived from this class without getting issues. Is that related to #63 ?

templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/ApplicationComponentTest.java
maybeec: You are definitively right. Sry, could not see this comment in my app :/

hohwille:
You sill did not answer my question. I have no issue merging this (and hope it wont break any tests of existing projects what might be a little tricky but they can simply get the old behaviour back via their own test superclass if needed).

hohwille:
To proceed, I will merge and we clarify the questions in #60

Issue #12, opened 2018-09-24
Upgrade to spring-boot 2.1.x with hibernate 5.3 and JPA 2.2


hohwille:
Hibernate 5.3 is already out since a while and fully supports JPA 2.2:
https://www.thoughts-on-java.org/hibernate-5-3/
We should therefore use the official JPA 2.2 by replacing
```
      <dependency>
        <groupId>org.hibernate.javax.persistence</groupId>
        <artifactId>hibernate-jpa-2.1-api</artifactId>
        <version>1.0.0.Final</version>
      </dependency>
```
with
```
      <dependency>
        <groupId>javax.persistence</groupId>
        <artifactId>javax.persistence-api</artifactId>
        <version>2.2</version>
      </dependency>
```
To do so we also need to upgrade to hibernate 5.3 what will come with spring-boot 2.1.0:
https://github.com/spring-projects/spring-boot/issues/11725
But currently not even the milestone 2.1.0.M1 is in maven central.

maybeec:
SpringBoot 2.1 is release since a while and it comes with Java 11 support.
Currently, Java 11 support is most valuable for projects as Java 8 is at EOL and Java 11 will be the next LTS to go for. Therefore, I think we should definitively do the migration soon.

anmaso:
I also think we should go evergreen JVM and give support to older JVMs as exceptional cases.

maybeec:
Nicely, SpringBoot 2.1 did not remove support for Java8, so we finally could and should have support for Java LTS versions.

maybeec:
I just did it for my project, which seems to work smoothlessly.
Anyhow, we are in a early phase, so maybe some more bigger applications might face some issues. For my project I could not find any reason, why it should not work from the upgrade documentation.

https://github.com/spring-projects/spring-booT/wiki/Spring-Boot-2.1-Release-Notes#upgrading-from-spring-boot-20

So possibly low effort for devonfw.

hohwille:
It would be lovely if we also add a migration step for devcon. Ideally from devon4j:3.0.0 to 3.0.1 (with only the version update) and from there to 3.1.0 containing all required changes we might see for this upgrade here.

hohwille:
Done.

hohwille:
I just tested these changes and found a severe issue:
```
[...]

PR #63, opened 2019-02-28
Spring boot upgrade 2.1.2.RELEASE

AbhayChandel:
This PR is for issue https://github.com/devonfw/devon4j/issues/12

### DIFF
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -89,10 +89,6 @@
         <type>pom</type>
       </dependency>
       <dependency>
-        <groupId>javax.persistence</groupId>
-        <artifactId>javax.persistence-api</artifactId>
-      </dependency>
-      <dependency>
         <groupId>cglib</groupId>
         <artifactId>cglib</artifactId>
         <version>3.2.5</version>
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -88,19 +88,10 @@
         <version>1.3</version>
         <type>pom</type>
       </dependency>
-      <!-- Java Persistence API (2.x not available in groupId javax.persistence) -->
       <dependency>
-        <groupId>org.hibernate.javax.persistence</groupId>
-        <artifactId>hibernate-jpa-2.1-api</artifactId>
-        <version>1.0.0.Final</version>
-      </dependency>
-      <!--
-        <dependency>
         <groupId>javax.persistence</groupId>
         <artifactId>javax.persistence-api</artifactId>
-        <version>2.2</version>
-        </dependency>
-      -->
+      </dependency>
       <dependency>
         <groupId>cglib</groupId>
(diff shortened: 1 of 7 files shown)


boms/bom/pom.xml
```diff
@@ -88,19 +88,10 @@
...
       <dependency>
-        <groupId>org.hibernate.javax.persistence</groupId>
-        <artifactId>hibernate-jpa-2.1-api</artifactId>
-        <version>1.0.0.Final</version>
-      </dependency>
-      <!--
-        <dependency>
         <groupId>javax.persistence</groupId>
         <artifactId>javax.persistence-api</artifactId>
-        <version>2.2</version>
```
hohwille: why did you remove the version? IMHO this is a mistake. A dependency in depencenyManagement without a version seems pointless to me. Or is this a miss-interpretation of the diffs?
https://github.com/devonfw/devon4j/blob/d9d7403493513a24a14b363d3701fbd67b6d72ce/boms/bom/pom.xml#L93

hohwille: If that dependency comes in 2.2+ version via springs BOMs and we do not want to repeat it here, you should remove it entirely.

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
```diff
@@ -172,7 +173,7 @@ public void execute(Operation operation, List<String> configurations, String job
     SpringApplication app = new SpringApplication(configurationClasses);
 
     // no (web) server needed
-    app.setWebEnvironment(false);
+    app.setWebApplicationType(WebApplicationType.NONE);
```
hohwille: Fine fix. However, this reminds me of the fact that our batch still needs rework and polising as this is old fashion style code. But removing the spring XMLs and moving to a modern spring-boot only config is not in the scope of this PR, so all perfectly right.

boms/bom/pom.xml
AbhayChandel: Nice catch @hohwille , javax.persistence-api:2.2 is already defined in spring-boot-depenedencies:2.1.2.RELEASE. So I think it is better to remove
it completely from devon4j-bom. Removed and changes pushed to the PR.

AbhayChandel:
> @AbhayChandel Thanks. Looking fine now. As we rely on an external dependencyManagement then, can you confirm this:
> 
> * we do not depend on this proprietary dependency anymore `org.hibernate.javax.persistence:hibernate-jpa-2.1-api`?
> * instead we depend on `javax.persistence:javax.persistence-api:2.2`, correct?

Yes, we are no more depending devon4j on `org.hibernate.javax.persistence:hibernate-jpa-2.1-api?`. It has been replaced with `javax.persistence:javax.persistence-api:2.2`.

PR #64, opened 2019-02-28
JAVA 8 New Date/Time API- mtsj #152

vapadwal:
This issue is related to Java Date/Time Api.

please see review comments for PR in mtsj https://github.com/devonfw/my-thai-star/pull/192


templates/server/src/main/resources/archetype-resources/core/src/main/resources/config/app/common/dozer-mapping.xml
```diff
@@ -12,6 +12,19 @@
...
+      <copy-by-reference>java.time.ZonedDateTime</copy-by-reference>
+      <copy-by-reference>java.time.Instant</copy-by-reference>
+      <copy-by-reference>java.time.Year</copy-by-reference>
+      <copy-by-reference>java.time.Duration</copy-by-reference>
+      <copy-by-reference>java.time.Clock</copy-by-reference>
+      <copy-by-reference>java.time.MonthDay</copy-by-reference>
+      <copy-by-reference>java.time.OffsetDateTime</copy-by-reference>
+      <copy-by-reference>java.time.OffsetTime</copy-by-reference>
+      <copy-by-reference>java.time.Period</copy-by-reference>
+      <copy-by-reference>java.time.YearMonth</copy-by-reference>
```
hohwille: Excellent. You also cared about dozer. Great. The list looks quite good. I am missing `java.time.Month` however.

modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java
```diff
@@ -136,6 +140,11 @@ public ObjectMapper createInstance() {
       mapper.setSubtypeResolver(subtypeResolver);
     }
 
+    mapper.setSerializationInclusion(Include.NON_NULL);
```
hohwille: Can we add a comment above:
```
// omit properies in JSON that are null
```

modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java
```diff
@@ -136,6 +140,11 @@ public ObjectMapper createInstance() {
...
+    mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
+    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
hohwille: Can we add a comment above:
```
// ignore unkown properties in JSON to prevent errors
// e.g. when the service has been updated/extended but the calling REST client is not yet updated
// see https://github.com/devonfw-wiki/devon4j/wiki/guide-service-layer#versioning

File: modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java [hunk 4]
```diff
@@ -136,6 +140,11 @@ public ObjectMapper createInstance() {
       mapper.setSubtypeResolver(subtypeResolver);
     }
 
+    mapper.setSerializationInclusion(Include.NON_NULL);
+    mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```
hohwille: We should at least add a comment above:
```
// Write legacy date/calendar as readable text instead of numeric value
// See https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_DATES_AS_TIMESTAMPS
```

This is the only line that could be questioned. We could also remove this setting. However, java.util.Date/Calendar is legacy crap anyhow.

modules/json/pom.xml
```diff
@@ -33,6 +33,18 @@
       <artifactId>spring-data-commons</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.module</groupId>
+      <artifactId>jackson-module-parameter-names</artifactId>
```
hohwille: Can you give any rationale why we need this dependency?
I do not understand how this is related and have never used this one before.

modules/json/pom.xml
```diff
@@ -33,6 +33,18 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jdk8</artifactId>
```
hohwille: this one is IMHO also not required for this PR. It IMHO adds support for optinals.
Maybe we should leave it optional if someone wants support for optionals. For our modules we should keep dependencies as minimal as possible.

hohwille:
Finally I do have a further important suggestion for improvement. Move these configuration code to another proected method. If somebody wants to change it he otherwise would not have any option to do so.

We could also move this change to our app template...

templates/server/src/main/resources/archetype-resources/core/src/main/resources/config/app/common/dozer-mapping.xml
hohwille: Not obviously in the diffs but done:
https://github.com/vapadwal/devon4j/blob/issue_javatime8/templates/server/src/main/resources/archetype-resources/core/src/main/resources/config/app/common/dozer-mapping.xml#L28
👍 Thanks @vapadwal

modules/json/pom.xml
hohwille: @vapadwal 👍 Thanks for updating this. Done:
https://github.com/vapadwal/devon4j/blob/issue_javatime8/modules/json/pom.xml#L38

modules/json/pom.xml
hohwille: Done:
https://github.com/vapadwal/devon4j/blob/issue_javatime8/modules/json/pom.xml#L38

hohwille:
@vapadwal Thanks for your update on this PR. You resolved the issues. 👍 
Sorry, but I still have one more suggestions. Your new way to configure Jackson bean mapper seems to introduce some indirections that are IMHO unecessary complex. Also you named them `configPropertiesJavaTime` even though the actual options are not releated to Java tome. Instead of defining our own properties as Strings as well as as Enum that 1:1 represent Jackson properties I would suggest:

We keep the Java8 Module as default in the json module (as it currently is):
```
mapper.registerModule(new JavaTimeModule());
```
But remove the `configPropertiesJavaTime` with the according if block and the new constants, etc.
Instead we move the other 3 config lines to the archetype (https://github.com/vapadwal/devon4j/blob/issue_javatime8/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/config/ApplicationObjectMapperFactory.java) but simply directly via Jackson API without any shim in between:
```
mapper.setSerializationInclusion(Include.NON_NULL);
mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```

modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java
```diff
@@ -31,6 +34,12 @@
 
   private SimpleModule extensionModule;
 
+  private List<String> configPropertiesJavaTime;
+
+  private static final String INCLUDE_NON_NULL = "Include_NON_NULL";
+
+  private static final String DESERIALIZATIONFEATURE_FAIL_ON_UNKNOWN_PROPERTIES = "DeserializationFeature_FAIL_ON_UNKNOWN_PROPERTIES";
```
hohwille: Please remove these 3 fields.

modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java
```diff
@@ -135,8 +144,34 @@ public ObjectMapper createInstance() {
...
+            break;
+          case DESERIALIZATIONFEATURE_FAIL_ON_UNKNOWN_PROPERTIES:
+            // ignore unknown properties in JSON to prevent errors
+            // e.g. when the service has been updated/extended but the calling REST client is not yet updated
+            // see https://github.com/devonfw-wiki/devon4j/wiki/guide-service-layer#versioning
+            mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, true);
+            break;
+        }
+      }
+    }
```
hohwille: Please remove this if block.

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/config/ApplicationObjectMapperFactory.java
```diff
@@ -29,5 +35,7 @@ public ApplicationObjectMapperFactory() {
 	// register spring-data Pageable
     module.addSerializer(Pageable.class, new PageableJsonSerializer());
     module.addDeserializer(Pageable.class, new PageableJsonDeserializer());
+    // configure properties for Java Time module
+    setConfigPropertiesJavaTime(this.configProps);
```
hohwille: Please remove and instead override `createInstance()` method using `super.createInstance()` and tweak the result with the 3 suggested options. This way we do not change the behaviour of existing projects (potentially breaking anything) and only offer these configs as suggested for new projects.

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/config/ConfigJavaTimeProperties.java
```diff
@@ -0,0 +1,15 @@
...
+public enum ConfigJavaTimeProperties {
+  /**
+   * Enum for omit properies in JSON that are null
+   */
+  Include_NON_NULL,
+  /**
+   * Enum for ignoring unknown properties in JSON to prevent errors
+   */
+  DeserializationFeature_FAIL_ON_UNKNOWN_PROPERTIES
+}
```
hohwille: Please delete this enum.

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/config/ApplicationObjectMapperFactory.java
```diff
@@ -29,5 +32,17 @@ public ApplicationObjectMapperFactory() {
 	// register spring-data Pageable
     module.addSerializer(Pageable.class, new PageableJsonSerializer());
     module.addDeserializer(Pageable.class, new PageableJsonDeserializer());
+
+    ObjectMapper objectMapper = super.createInstance();
```
hohwille: @vapadwal almost perfect now, but this block needs to go to the method `createInstance` that is supposed to be overridden with this logic.

Issue #74, opened 2019-04-11
Maven version placeholders (CI-friendly-maven for devon4j)


AbhayChandel:
This will allow to build project with different versions without need to change the pom.xml. This is a very helpful feature especially for CI/CD environment.

To read more about this Maven feature follow link below.
https://maven.apache.org/maven-ci-friendly.html

hohwille:
I am already using this in devon-ide:
* https://github.com/devonfw/devon-ide/blob/master/.mvn/maven.config
* https://github.com/devonfw/devon-ide/blob/master/scripts/pom.xml#L14

hohwille:
Done for devon4j itself. We did not provide it within the archetype as the maven-archetype mechanism does not fit easily with this.

hohwille:
I have created #85 for the archetype and documentation. For `3.1.0` we close this issue and consider it feature complete.

PR #75, opened 2019-04-11
Added Maven version placeholders

AbhayChandel:
This PR is for issue https://github.com/devonfw/devon4j/issues/74

With this PR, Maven CI friendly versions can now be used for devon4j project and devon4j archetype template projects.


templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -10,6 +10,9 @@
   <description>Application based on the Open Application Standard Platform for Java (devon4j).</description>
 
   <properties>
+    <revision>0.0.1</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <projectVersion>${revision}${changelist}</projectVersion>
```
hohwille: It is good that we also provide this as best practice to our projects via archetype. However some concerns:
* you are so fat just setting a property `projectVersion` but this is not related to the actual `${project.version}` of that maven project. IMHO you need to change this in the `version` tag and we have to test how `maven-archetype` mechanism works with this as it is usually setting a concrete version in the `pom.xml` files that can be provided on the commandline on archetype generation
* many people will never have heared about this feature and are clueless about the benefits and how to use it properly. We should therefore create a small guide in devon4j and link the ci-friendly maven documentation from there.
* then we should add an XML comment to that guide here

templates/server/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -6,6 +6,9 @@
   <id>devon4j-template-server</id>
 
   <requiredProperties>
+    <requiredProperty key="version">
+            <defaultValue>${projectVersion}</defaultValue>
```
hohwille: your indendation seems to be broken. What IDE are you using? should be 2 spaces per indendation level (and not tabs)

pom.xml
```diff
@@ -22,11 +22,14 @@
     <java.version>1.8</java.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <devon4j.version>3.1.0-SNAPSHOT</devon4j.version>
+    <revision>3.1.0</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <devon4j.version>${revision}${changelist}</devon4j.version>
```
hohwille: in devon-ide I decided to keep the actual version in `.mvn/maven.config` that I added to the git repo. This allows to externalize the version information out of all `pom.xml` files and reduce complexity for release creation (version bumping, setting next SNAPSHOT version). I also set the revision to `dev` in the POM itself to avoid redundancy and to enforce that if the `.mvn/maven.config` got lost, we have an abvious impact and do not accidentially create a wrong release-number (looking like a normal release). I would recommend to do the same here espcially as the redundant values will either lead to maintaining both places in parallel or to cause confusion if they run out of sync.

templates/server/src/main/resources/META-INF/maven/archetype-metadata.xml
hohwille: Where is the `version` variable used? I do not entirely get this idea.

templates/server/src/main/resources/archetype-resources/pom.xml
hohwille: BTW: To get started, you can just create an almost empty guide in the wiki and add the CI-friendly link and a TODO. We can update the wiki page later. Also I can contribute to that as well. For the comment use the link to the official wiki instead of the `devonfw-wiki` one.

AbhayChandel: Wiki won't be required as of now. As the CI friendly maven versions for archetype will not be implemented as part of this feature.

templates/server/src/main/resources/META-INF/maven/archetype-metadata.xml
AbhayChandel: This part has been removed now.

Issue #77, opened 2019-04-29
Remove devon4j-configuration module


hohwille:
devon4j shipps with a module called `configuration`. However, this seems to be empty and hence pointless. Can we just remove this entirely?

hohwille:
This module is a leftover from `oasp4j`:
https://github.com/oasp/oasp4j/blob/b06c19cd1d6a98b7d8836bd54342ab2c7178175a/modules/configuration/src/main/java/io/oasp/module/configuration/common/api/ApplicationConfigurationConstants.java

It already became obsolete when we introduced using spring-boot. For compatibility issues of oasp4j we keept this as empty module. With the creation of `devon4j` we should have removed it. So we are safe to remove it now in `devon4j` and can add a migration step for devcon/devon-ide to assure the dependency is removed if still present in some project.

hohwille:
@AbhayChandel or @vapadwal can somebody of you create a PR that is just deleting this module?

hohwille:
Done.

PR #83, opened 2019-05-03
#77 removed configuration

hohwille:
also removed docs folder (OOMPH)


hohwille:
@maybeec As I also kill OOMPH with this PR could you please give us feedback if that is still required somewhere? I never understood why it was added to oasp4j/devon4j and not to our ide?

maybeec:
Fine for me, but you should as well remove the reference to this oomph config file in devon-ide configuration.

hohwille:
Thanks for the feedback. devon-ide is completely renewed. oasp4j-ide is not maintained anymore.

Issue #72, opened 2019-03-28
Migrate to newest version of dozer


maybeec:
The development of dozer has been restarted.
If we are not going for other mapping tools by default, we should at least migrate to the latest version of dozer fixing multiple often occurring issues in bigger projects.

https://github.com/DozerMapper/dozer

hohwille:
@AbhayChandel please comment the current status. If done then also assign milestone and close.

hohwille:
I got the response via mail:

> Dozer is not managed by Spring boot. Hence we need to update it at our end explicitly.
> In devon4j this is being managed using devon4j-bom and the current version is 5.5.1.
> Not sure if it will be straightforward to upgrade to the latest version 6.4.1, since it is different project then what it was back in 2014.
> I suggest, we can take up the migration and if the PR is stable enough then we can include it in 3.1.0.

Agreed. Lets start with PR.

hohwille:
Thanks to @Rohit90700 this is fixed via PR #86

hohwille:
I always thought we never made our fundamental features work with orika.
However, I just saw that we did since long:
* https://github.com/devonfw/devon4j/blob/develop/modules/beanmapping/src/main/java/com/devonfw/module/beanmapping/common/impl/orika/CustomMapperEto.java
* https://github.com/devonfw/devon4j/blob/develop/modules/beanmapping/src/test/java/com/devonfw/module/beanmapping/common/impl/orika/BeanMapperImplOrikaTest.java

So if we are planning an incompatible change for beanmapping anyhow with `3.2.0` why dont we go and switch to orika officially? Aint this about time?
* https://www.baeldung.com/java-performance-mapping-frameworks
* https://github.com/oasp/oasp4j/issues/4

maybeec:
just read https://www.baeldung.com/java-performance-mapping-frameworks, I think it might be interesting to think about MapStruct as the default in the meantime.

hohwille:
@maybeec MapStruct is not a drop-in replacement. It is a code-generater and can not serve with a generic API to map arbitrary objects. IMHO the philosophy is so different that it is not really to be considered for our existing BeanMapper API.
See https://github.com/oasp/oasp4j/issues/4

hohwille:
I created issue #147 for this and therefore close this issue as it has already been done as planned initially.

hohwille:
For the record:
https://github.com/devonfw/devon4j/commit/cadadf5f3191d8af36663d42c592296ce5afb8f6
[...]

PR #86, opened 2019-05-08
Dozer Mapper Upgrade

Rohit90700:
Dozer Upgraded with the latest version 6.4.1 (Issue:- https://github.com/devonfw/devon4j/issues/72)

### DIFF
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -166,11 +166,12 @@
         <artifactId>json</artifactId>
         <version>20180130</version>
       </dependency>
-      <!-- Dozer for bean-mapping/conversion (Alternative to Orika) -->
+      <!-- Dozer for bean-mapping/conversion -->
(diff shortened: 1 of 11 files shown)


hohwille:
In dozer namespaces changed entirely. Great thing as that we provide a shim with BeanMapper from devon4j that is hiding these changes. However, the dozer XML configuration namespace also needs to change. Therefore we should also extend the devon4j migration logic here:
https://github.com/devonfw/devon-ide/pull/108

Rohit90700:
> In dozer namespaces changed entirely. Great thing as that we provide a shim with BeanMapper from devon4j that is hiding these changes. However, the dozer XML configuration namespace also needs to change. Therefore we should also extend the devon4j migration logic here:
> [devonfw/devon-ide#108](https://github.com/devonfw/devon-ide/pull/108)

Hi @hohwille , I have updated your PR https://github.com/devonfw/devon-ide/pull/108 with the dozer mapper changes.

Issue #67, opened 2019-03-12
Batch template cannot be started


maybeec:
It seems, that the existing batch configuration served by the archetype is not working at all. First there is an exception as it tries to create a security bean although global method security is disabled.
Enabling it (why is it disabled?), you get an exception, that jobRegistry Bean is not provided.

I am using oasp 3.0.0, but I did not detect any difference in the latest template.

Also there seem to be EndpointAutoCofiguration excluded, but JmxAutoConfiguration not, which is not working either.

Just remembered, meanwhile I upgraded to latest SpringBoot Release. So maybe these issues are also as of this upgrade. Anyhow, maybe a good chance to check again.

hohwille:
@maybeec can you give us some more hints to reproduce the problem?
We do have automated tests of the archetype including the batch option.
So does your error occur only when you start the spring boot app for the batch manually?
Otherwise if this is only related to your project using a more recent spring-boot version this issue will be considered as invalid but we can join forces to create version `3.1.0` such that it suits your needs.

maihacke:
Could you please show here how exactly did you try to start the batch?

maybeec:
This was referring to oasp4j 3.0.0, maybe this has been resolved already. Anyhow, I was generating the batch application by the archetype and starting the batch generated spring boot application without any success.

maihacke:
Did you try this from eclipse or via cli?
At least for me it looks like our documentation is broken. The examples for starting the batches cannot work in opinion - at least the classpath is missing...

maybeec:
I just tried by eclipse, `SpringBootApp -> Run as Java` these times.

hohwille:
With the merged PR it is possible to start the batch from the template/archetype.
I close this issue. If you still have issues with batch, create a new concrete issue.

PR #89, opened 2019-05-16
fix batch template issue #https://github.com/devonfw/devon4j/issues/67

vapadwal:
Fix batch template issue https://github.com/devonfw/devon4j/issues/67
Also removed the reference of devon4j-configuration from jpa-basic pom


modules/jpa-basic/pom.xml
```diff
@@ -59,11 +59,6 @@
       <artifactId>devon4j-test</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.devonfw.java.modules</groupId>
-      <artifactId>devon4j-configuration</artifactId>
-      <scope>test</scope>
-    </dependency>
```
hohwille: ouch. That was a bug I introduced with PR #83 
Interesting that travis and our batch archetype IT test did not reveal it.
Thanks for fixing.

templates/server/src/main/resources/archetype-resources/__batch__/src/main/java/__packageInPathFormat__/SpringBootBatchApp.java
```diff
@@ -13,10 +13,8 @@
 /**
  * {@link SpringBootApplication} for running this batch.
  */
-@SpringBootApplication(exclude = { EndpointAutoConfiguration.class, SecurityAutoConfiguration.class,
-SecurityFilterAutoConfiguration.class, })
+@SpringBootApplication
 @EntityScan(basePackages = { "${package}" }, basePackageClasses = { AdvancedRevisionEntity.class })
-@EnableGlobalMethodSecurity(jsr250Enabled = false)
```
hohwille: So we are disabling autorization in the batch?
Usually in my projects we did a login from the batch launcher.
If that is currently not present at least better to have a batch that has all permissions than none ;)

Issue #98, opened 2019-06-03
Add equalsVerifier to archetype


hohwille:
Add this dependency to `api` module:
```xml
    <dependency>
      <groupId>nl.jqno.equalsverifier</groupId>
      <artifactId>equalsverifier</artifactId>
      <scope>test</scope>
    </dependency>
```

And this test:
```
public class ToTest extends ModuleTest {

  private static final ToExclusions EXCLUSIONS = new ToExclusions();

  static {
    // EXCLUSIONS.add(MyBusinessEto.class).ignoreEquals().ignoreProperties("foo", "bar");
  }

  /**
   * Finds all {@link AbstractTo Transportobjekte} of this app via reflection and tests all getters/setters as well as equals and hashCode.
   */
  @Test
  public void testTos() {

    List<Class<?>> toClasses = getToClasses();
    SoftAssertions assertion = new SoftAssertions();
    for (Class<?> toClass : toClasses) {
      ToExclusion exclusion = EXCLUSIONS.get(toClass.getName());
      testGetSet(toClass, assertion, exclusion);
      testEqualsAndHashcode(toClass, assertion, exclusion);
    }
    assertion.assertAll();
  }

  private void testEqualsAndHashcode(Class<?> clazz, SoftAssertions assertion, ToExclusion exclusion) {

    if ((exclusion != null) && (exclusion.ignoreEquals)) {
      return;
    }
    EqualsVerifierApi<?> verifier = EqualsVerifier.forClass(clazz).withRedefinedSuperclass().usingGetClass().suppress(Warning.NONFINAL_FIELDS, Warning.INHERITED_DIRECTLY_FROM_OBJECT);
    try {
      verifier.verify();
    } catch (AssertionError e) {
      assertion.fail(e.getMessage(), e);
    }
  }

  private void testGetSet(Class<?> clazz, SoftAssertions assertion, ToExclusion exclusion) {

    if (!Modifier.isAbstract(clazz.getModifiers())) {
      ReflectionUtils.doWithLocalFields(clazz, field -> {
        try {
          if (isCheckable(clazz, field, exclusion)) {
            Object testInstance = clazz.newInstance();
            Method getMethod = null;
            Method setMethod = null;

            if (Boolean.class.equals(field.getType()) || boolean.class.equals(field.getType())) {
              getMethod = findMethod("is", clazz, field);
              if (getMethod == null) {
                getMethod = findMethod("get", clazz, field);
              }
            } else {
[...]

PR #103, opened 2019-06-12
#98: Added equalsVerifier dependency and tests for transfer objects

lmarniazman:
PR for #98


hohwille:
@lmarniazman thanks for this PR.
Have you seen that you do have caused compile errors?
```
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[19,28] package org.assertj.core.api does not exist
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[27,43] package com.devonfw.module.test.common.base does not exist
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[33,29] cannot find symbol
[INFO]   symbol: class ModuleTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[57,54] cannot find symbol
[INFO]   symbol:   class SoftAssertions
[INFO]   location: class it.pkg.ToTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[70,43] cannot find symbol
[INFO]   symbol:   class SoftAssertions
[INFO]   location: class it.pkg.ToTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[202,44] cannot find symbol
[INFO]   symbol:   class AbstractTo
[INFO]   location: class it.pkg.ToTest.ToExclusions
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[48,5] cannot find symbol
[INFO]   symbol:   class SoftAssertions
[INFO]   location: class it.pkg.ToTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[48,36] cannot find symbol
[INFO]   symbol:   class SoftAssertions
[INFO]   location: class it.pkg.ToTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[175,56] cannot find symbol
[INFO]   symbol:   class AbstractTo
[INFO]   location: class it.pkg.ToTest
[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/enterprise/project/enterprise/api/src/test/java/it/pkg/ToTest.java:[186,9] cannot find symbol
[INFO]   symbol:   method fail(java.lang.String)
[INFO]   location: class it.pkg.ToTest
```
Will you be able to fix this and update this PR by pushing these fixing commits to your fork?

hohwille:
@lmarniazman any progress on this PR? I thought we solved most of the problems together. Devon `3.1.0` release is comming very close and we are likely to miss it...

lmarniazman:
@hohwille sorry for the delay, on the next startup of my eclipse IDE I got the same error as the one we fixed last time. I haven't got around to taking a closer look on it, but I will now.

lmarniazman:
I added two additional warning suppressors in the last commit as a workaround for the EqualsVerifier errors thrown by BinaryObjectEto and UserProfileTo. 
#98_equalsVerifier_errors
I will try to fix them in another commit.

lmarniazman:
In the last commit, I deleted the warning suppressors and implemented a method that checks if the equals() or hashCode() method in a TO has been overridden. Only in this case, testEqualsAndHashCode() of ToTest is run on it.

After talking to Malte about how to proceed with the suppressed warnings, we thought this would be a good solution, because the ToTest doesn't necessarily have to run on classes that don't have their own implementation of equals() and hashCode().

What are your thoughts on this @hohwille?

templates/server/src/main/resources/archetype-resources/api/src/test/java/__packageInPathFormat__/ToTest.java
```diff
@@ -55,12 +55,30 @@ public void testTos() {
...
+
+	boolean existing = false;
+
+	try {
+	  if(clazz.getMethod("equals", Object.class).getAnnotation(Override.class) != null) existing = true;
+	} catch(NoSuchMethodException noEquals) { existing = false; }
+
+	try {
+	  if(clazz.getMethod("hashCode", null).getAnnotation(Override.class) != null) existing = true;
+	} catch(NoSuchMethodException noHashCode) { existing = false; }
```
maybeec: 1. equals and hashCode of class Object is always there. So you have to ask for it of the clazz object
2. don't check for the Override annotation. It's an optional annotation and therefore should not be used to determine the existence.
3. We should test for both methods separately if possible. Just in case someone is just implementing an equals function and not a hashcode.

lmarniazman: Number 2 and 3 are good suggestions, I didn't catch those mistakes yesterday. 

I'm not sure what you mean by number 1 though. The method checks for equals() and hashCode() of the clazz object. The 'Object.class' parameter only specifies the parameter of the equals() method, so that the method signatures can be compared.

lmarniazman: The new implementation checks for both methods seperately, but runs the test only if both methods were overridden. I'll look into a way to make the verifier only test one of each if not both methods were implemented.

templates/server/src/main/resources/archetype-resources/api/src/test/java/__packageInPathFormat__/ToTest.java
```diff
@@ -58,32 +60,30 @@ public void testTos() {
...
+	  if(method.toString().equals("public native int java.lang.Object.hashCode()")) return false;
+	}
 
 	return true;
   }
 
   private void testEqualsAndHashcode(Class<?> clazz, SoftAssertions assertion, ToExclusion exclusion) {
 
-	if(checkForEquals(clazz) || checkForHashCode(clazz)) {
+	if((checkForCustomEquals(clazz) == false) || (checkForCustomHashCode(clazz) == false)) {
```
maybeec: in general you would do it as !checkForCustomEquals(clazz) rather than comparing the equality.
Also it would be easier to read if you provide a more fluent readable name for the methods. Thus, you would not even had made the error initially with the && ;)
For example call the method `hasCustomEquals`, then the return value is clearly understood.

hohwille:
> because the ToTest doesn't necessarily have to run on classes that don't have their own implementation of equals() and hashCode().

I do not get this. We are only checking TOs. These derive from AbstractTo what already overrides equals and hashCode. Technically there is no way for a TO not to override equals and hashCode as otherwise their implementation is entirely broken.

templates/server/src/main/resources/archetype-resources/api/pom.xml
```diff
@@ -29,11 +29,24 @@
...
+    <dependency>
+      <groupId>com.devonfw.java.modules</groupId>
+      <artifactId>devon4j-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- EqualsVerifier -->
+    <dependency>
+      <groupId>nl.jqno.equalsverifier</groupId>
+      <artifactId>equalsverifier</artifactId>
+      <version>3.1.9</version>
```
hohwille: As we are proposing this as the default/standard for devonfw projects, we should even better add this dependency to `devon4j-test` directly so all projects get it from there implicitly.

templates/server/src/main/resources/archetype-resources/api/src/main/java/__packageInPathFormat__/general/common/api/to/UserProfileTo.java
```diff
@@ -34,4 +35,10 @@ public void setLogin(String login) {
     this.login = login;
   }
 
+  @Override
+  protected void toString(StringBuilder buffer) {
+
+	buffer.append(getClass().getSimpleName() + ": " + "login = " + login + ", hashcode = " + this.hashCode() + ";");
```
hohwille: please check your IDE. Seems you are not using Eclipse with proper formatter and save actions.

templates/server/src/main/resources/archetype-resources/api/src/test/java/__packageInPathFormat__/ToTest.java
```diff
@@ -0,0 +1,295 @@
...
+    assertion.assertAll();
+  }
+
+  /**
+   * Checks for an overridden equals() method of a given class object
+   */
+  private boolean checkForCustomEquals(Class<?> clazz) {
+
+	for(Method method : clazz.getMethods()) {
+	  if(method.toString().equals("public boolean java.lang.Object.equals(java.lang.Object)")) return false;
```
hohwille: simply use `clazz.getDeclaredMethod("equals", new Class[] { Object.class})` instead.
BTW: Your formatter is entirely broken. We are trying to enforce standard formatting rules so that we do not end up in git diff wars when people do changes and reformat code. Please fix your IDE to be compliant with devonfw.

lmarniazman: Thanks for the comments. I'll add the dependency to the test module. Should I still fix the methods that check for custom equals() / hashCode() implementations? As far as I can see, AbstractTo only overrides the toString() method, so a class not overriding equals() and hashCode() is possible.

templates/server/src/main/resources/archetype-resources/api/pom.xml
lmarniazman: Thanks for the comment. I'll add the dependency to the test module.

lmarniazman:
I'm not sure if this was the right location to add the dependency. Pointers are appreciated.

templates/server/src/main/resources/archetype-resources/api/src/main/java/__packageInPathFormat__/general/common/api/to/UserProfileTo.java
hohwille: In case you are still struggling with the IDE, then simply add two whitespaces to these 2 `toString` methods you have added.
BTW: Why did you add the hashCode to the toString result?

hohwille:
@maybeec github requires an offiical confirmation from your end that your review changes are satisfied. As your review comment was resolved quite some time ago, I am going to merge this anyway. 

Still there are minimal formatting/indendation issues and one toString that still appends hashCode() but this is nice to have. So I merge this and fix the cosmetic issues after the merge.

PR #110, opened 2019-07-01
Update link to devon4j-doc-3.0.0.pdf

markusschuh:


hohwille:
@markusschuh excellent catch. Thanks for the PR and your contribution.

Issue #108, opened 2019-06-24
javadoc not working with openjdk 11


hohwille:
We fail to build our releases with openjdk11 due to this issue:
https://bugs.openjdk.java.net/browse/JDK-8212233

See also #16 

So to build releases of devon4j we either need to stick to jdk 8 or go to 12 (not yet testet) or use oraclejdk 11.

hohwille:
It is working with 11.0.4 (that we use from AdoptJDK). However, some additional config in maven is still required before we can close the release.

hohwille:
This issue caused the release to fail:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.0:jar (attach-javadocs) on project devon4j-test: Execution attach-javadocs of goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.0:jar failed.: NullPointerException -> [Help 1]
```
Seems as if first the javadoc plugin needs to be updated.

hohwille:
We need `3.1.0` as already documented here:
https://github.com/devonfw/devon4j/blob/develop/documentation/guide-jdk.asciidoc#maven-plugins

hohwille:
Further, we need to set source to 8:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8212233
As already done for devon-ide:
https://github.com/devonfw/devon-ide/blob/85d599c8bf8824904a6cc1ee6279bc7bad40c24d/pom.xml#L149

hohwille:
So the required changes to our POM are done with PR #140.
We further switched to AdoptOpenJDK in `devon-ide` so we get jdk `11.0.4` that contains the necessary fixes for javadoc tool.

Issue #97, opened 2019-06-03
dependency-check-maven should not break site build


hohwille:
We have configured `dependency-check-maven` as a build plugin that can be enabled via a profle (`security`) as well as as a report for the maven site.
The latter does break the site generation. This is entirely odd and non-sense (IMHO already from the OWASP plugin itself) as the site should give reporting of the current state. It also can be build if tests fail and lists the failing tests, etc. The same is to be expected for CVEs.
With a build breaker also for the site other important reports can not be generated what sucks.
We need to configure an option that prevents the build to break.

hohwille:
https://github.com/devonfw/devon4j/blob/08e6ebd81d70c95c9212ca7287d0e9b48ed67fcf/templates/server/src/main/resources/archetype-resources/pom.xml#L470

hohwille:
* https://jeremylong.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#failBuildOnCVSS
* https://jeremylong.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#failOnError

hohwille:
Done with #120.

PR #113, opened 2019-07-03
Fixed build failure due to dependency-check-maven vulnerabilities

AbhayChandel:
This PR is for issue #97. Following changes were made as part of this PR:

- Build failure was disabled by setting _failOnError_ to _false_ and _failBuildOnCVSS_ was removed.
- _dependency-check-maven_ plugin was updated from 3.1.1 to 5.1.0.
- Cleaned up leftover from issue #77.

The same was tested with Java 11. For which :
- Was able to run _verify_ goal successfully with _dependency-check-maven_ plugin.
- However, _site_ goal failed due to error in _maven-javadoc-plugin_ causing _javadoc: error - Error fetching URL: (http://www.devonfw.com/devon4j/)_ . I am not sure but I think this is known issue and is related to issue #108.

### DIFF
--- a/boms/minimal/pom.xml
+++ b/boms/minimal/pom.xml
@@ -52,11 +52,6 @@
       </dependency>
       <dependency>
         <groupId>com.devonfw.java.modules</groupId>
-        <artifactId>devon4j-configuration</artifactId>
-        <version>${project.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>com.devonfw.java.modules</groupId>
         <artifactId>devon4j-security</artifactId>
         <version>${project.version}</version>
       </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -336,7 +336,7 @@
         <plugin>
           <groupId>org.owasp</groupId>
           <artifactId>dependency-check-maven</artifactId>
-          <version>3.1.1</version>
+          <version>5.1.0</version>
         </plugin>
         <plugin>
           <groupId>org.springframework.boot</groupId>
@@ -445,7 +445,7 @@
             <groupId>org.owasp</groupId>
             <artifactId>dependency-check-maven</artifactId>
             <configuration>
-              <failBuildOnCVSS>8</failBuildOnCVSS>
+              <failOnError>false</failOnError>
             </configuration>
             <executions>
               <execution>


boms/minimal/pom.xml
```diff
@@ -50,11 +50,6 @@
         <artifactId>devon4j-beanmapping</artifactId>
         <version>${project.version}</version>
       </dependency>
-      <dependency>
-        <groupId>com.devonfw.java.modules</groupId>
-        <artifactId>devon4j-configuration</artifactId>
```
hohwille: good catch. Seems I missed this one. 👍

pom.xml
```diff
@@ -336,7 +336,7 @@
         <plugin>
           <groupId>org.owasp</groupId>
           <artifactId>dependency-check-maven</artifactId>
-          <version>3.1.1</version>
+          <version>5.1.0</version>
```
hohwille: yep, latest is greatest 👍

pom.xml
```diff
@@ -445,7 +445,7 @@
             <groupId>org.owasp</groupId>
             <artifactId>dependency-check-maven</artifactId>
             <configuration>
-              <failBuildOnCVSS>8</failBuildOnCVSS>
+              <failOnError>false</failOnError>
```
hohwille: In general this is the config we need. However there are two ways to use the plugin:
1. as a single check (mvn dependency:check) [* Example 7]
2. as a report plugin for the maven site [* Example 2]

The site (2.) should always build. In the resulting HTML you can then see failing tests, vulnerabilities in depndencies, etc. whereas for (1.) the build-breaker still makes sense. Can you hence move the new config (failOnError) to the reporting config and keep the existing config for the build-section invocation (and CLI)?

(*) https://jeremylong.github.io/DependencyCheck/dependency-check-maven/

hohwille:
We missed the archetype which is IMHO even more important for this issue:

* https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/pom.xml#L269
* https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/pom.xml#L351
* https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/pom.xml#L465

hohwille:
@AbhayChandel thanks for the PR. I took your changes into #120 and #119.
As #97 is fixed we can also close this PR.

Issue #114, opened 2019-07-04
JSON mapping missing for IdRef


hohwille:
We have introduced `IdRef` but we are lacking JSON mapping supprt:
Solution from here:
https://github.com/devonfw-forge/keywi/blob/master/core/src/main/java/com/devonfw/keywi/general/common/impl/config/ApplicationObjectMapperFactory.java#L44
Should be added here:
https://github.com/devonfw/devon4j/blob/develop/modules/json/src/main/java/com/devonfw/module/json/common/base/ObjectMapperFactory.java#L162

Json Mappings come from here (should also be added to devon4j-json):
https://github.com/devonfw-forge/keywi/tree/master/core/src/main/java/com/devonfw/module/basic/common/api/reference/json

Also we need to update `dozer-mapping.xml` of archetype like here:
https://github.com/devonfw-forge/keywi/blob/master/core/src/main/resources/config/app/common/dozer-mapping.xml#L15
in:
https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/core/src/main/resources/config/app/common/dozer-mapping.xml

hohwille:
Ideally we could include this into `3.1.0`. Lets see as we are on the finishing line...

hohwille:
See #29 and #106

hohwille:
Done with PR #117

Issue #112, opened 2019-07-03
Add test to ObjectMapperFactory


vapadwal:
Add more tests (e.g. deserialization of Page, Pageable, Instant, Year, CsrfToken, etc.). 

Tests needed for ObjectMapperFactory

For more details refer to https://github.com/devonfw/devon4j/pull/111

hohwille:
Done with PR #117

PR #117, opened 2019-07-05
Feature/114 id ref json

hohwille:
Resolves:
* #114 - JSON Support for IdRef
* #112 - extended JUnit for OpenMapperFactory (JSON Mapping)
* still found bugs with JSON Mapping for spring-data pageing, included fixes
* also found a bug in an old config for ObjectMapper config, included fix


hohwille:
As we will have "code freeze" starting to morrow and we have to finalize our release, I will not wait for a review.

Issue #118, opened 2019-07-05
Avoid redundant versioning of devon4j itself and archetype


hohwille:
We have to maintain, update and fix versions often. In most cases we have trouble and issues because almost all versions are both in our devon4j parent POM as well as in the toplvel POM of the archetype.

* https://github.com/devonfw/devon4j/blob/develop/pom.xml#L22
* https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/pom.xml#L12

hohwille:
I see the following options:
* We use variables in the archetype like `<spring.boot.version>$[spring.boot.version]</spring.boot.version>` to avoid redundancies
* We really create a parent POM that also our archetype project will extend. This could be beneficial as we have the same redundancies also in various other repos such as:
  * https://github.com/devonfw/devon-ide/blob/master/pom.xml#L47
  *  https://github.com/devonfw/sonar-devon-plugin/blob/master/pom.xml#L128
  * https://github.com/devonfw-forge/keywi/blob/master/pom.xml#L64

hohwille:
IMHO all already done via first option.

PR #120, opened 2019-07-05
#97: updated owasp dependency-check and prevent breaking site build

hohwille:
#118: avoid redundant spring.boot.version
#112: renamed test to ObjectMapperFactoryTest


hohwille:
`mvn site:site` now succeeds.

PR #129, opened 2019-07-17
fix file master adding extension .asciidoc

Jorge-Dacal:


hohwille:
@Jorge-Dacal seems you need to sign CLA first.
@all: what is this master-devon4j all about? Why was it added? We alread have devon4-doc.asciidoc.

sjimenez77:
Hi @hohwille,

Sorry for the late reply, we needed a different master document that does not contain cover, TOC and other contents. We could have used devon4j-doc, but it would have been a bit weird to see another TOC and a different cover as the other sections.

Regards,
Santos

Issue #123, opened 2019-07-08
Update jacskon to 2.9.9.1


hohwille:
For #115 we need to update jackson to version `2.9.9.1`:
https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.9.9.1/

hohwille:
Done with PR #131

Issue #122, opened 2019-07-08
Update guava to version 28.0-jre


hohwille:
As part of #115 we need to update guava to `28.0-jre`:
https://repo1.maven.org/maven2/com/google/guava/guava/28.0-jre/

hohwille:
Done with PR #131

Issue #115, opened 2019-07-05
Resolve CVEs identified for 3.1.0 release


AbhayChandel:
Running _dependency-check-maven_ plugin as part of build identifies CVEs of score 8 and higher. Need to fix them as part of this issue.

Note: As part of this fix enable _failBuildOnCVSS_ property for build section.

hohwille:
Logback needs to be updated:
* https://nvd.nist.gov/vuln/detail/CVE-2017-5929

hohwille:
We are currently using logback classic `1.2.3`. This is the latest official version:
https://repo1.maven.org/maven2/ch/qos/logback/logback-classic/

Further, we are not using remote logging in devon4j so this is more or less a false-positive.

hohwille:
Here is the full list:
* [CVE-2019-12814](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2019-12814) from `jackson-databind-2.9.9.jar`
* [CVE-2018-10237](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2018-10237) from `guava-17.0.jar`
* [CVE-2019-3773](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2019-3773) from `spring-boot-starter-web-services-2.1.6.RELEASE.jar`
* [CVE-2018-1258](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2018-1258) from `spring-security-core-5.1.5.RELEASE.jar`
* [CVE-2019-3772](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2019-3772) from `spring-batch-integration-4.1.2.RELEASE.jar`
* [CVE-2017-5929](http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2017-5929) from `security-logging-logback-1.1.3.jar`

hohwille:
Here is the latest dependency tree of our archetype after merging PR #131:
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ basic-server ---
[INFO] archetype.it:basic-server:war:1.0.0-SNAPSHOT
[INFO] +- archetype.it:basic-core:jar:1.0.0-SNAPSHOT:compile
[INFO] |  +- archetype.it:basic-api:jar:1.0.0-SNAPSHOT:compile
[INFO] |  |  +- org.springframework.data:spring-data-commons:jar:2.1.9.RELEASE:compile
[...]

PR #131, opened 2019-07-30
Resolve CVEs identified

vapadwal:
Fix for issue https://github.com/devonfw/devon4j/issues/115

Fixed has been done for below issues 
CVE-2019-12814 from jackson-databind-2.9.9.jar - see issue #123
CVE-2018-10237 from guava-17.0.jar - see issue #122
CVE-2017-5929 from security-logging-logback-1.1.3.jar

Not able to find fix for below, as this Jars are already the latest ones.

CVE-2019-3773 from spring-boot-starter-web-services-2.1.6.RELEASE.jar
CVE-2018-1258 from spring-security-core-5.1.5.RELEASE.jar
CVE-2019-3772 from spring-batch-integration-4.1.2.RELEASE.jar

### DIFF
--- a/modules/json/pom.xml
+++ b/modules/json/pom.xml
@@ -18,6 +18,17 @@
     <dependency>
       <groupId>com.fasterxml.jackson.jaxrs</groupId>
       <artifactId>jackson-jaxrs-json-provider</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+    </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.9.9.1</version>
     </dependency>
     <dependency>
       <groupId>javax.inject</groupId>
@@ -36,6 +47,12 @@
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+    </exclusions>
     </dependency>
     <dependency>
       <groupId>com.devonfw.java.modules</groupId>
--- a/modules/json/pom.xml
+++ b/modules/json/pom.xml
@@ -18,17 +18,6 @@
     <dependency>
       <groupId>com.fasterxml.jackson.jaxrs</groupId>
       <artifactId>jackson-jaxrs-json-provider</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>com.fasterxml.jackson.core</groupId>
-          <artifactId>jackson-databind</artifactId>
-        </exclusion>
-    </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>2.9.9.1</version>
     </dependency>
     <dependency>
       <groupId>javax.inject</groupId>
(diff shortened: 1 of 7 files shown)


modules/json/pom.xml
```diff
@@ -18,6 +18,17 @@
...
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+    </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.9.9.1</version>
```
hohwille: Please consider putting versions into `dependencyManagement` of BOM.
If you need to exclude it and then include it, it seems as the `jackson-jaxrs-json-provider` is shipping with and old vulnerable version of it. Is there no new version of `jackson-jaxrs-json-provider` with the new dependency version to make it consistent? IMHO we should simply update/override `jackson.version` in our BOM.

modules/json/pom.xml
```diff
@@ -36,6 +47,12 @@
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
```
hohwille: if we succeed with my above suggestions, this can also be stripped and we get rid of such bells and whistles that are bad for maintenance.

modules/rest/pom.xml
```diff
@@ -18,6 +18,17 @@
...
+        <exclusion>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.9.9.1</version>
```
hohwille: Same here. The BOM is there to avoid such redundancies.

modules/json/pom.xml
vapadwal: I have found the latest version of bom for Jackson bom (2.10.0.pr1), they have just publish it. This jackson version excludes the vulnerable version of jar. I tried to update the Jackson.version in bom but it was not able to override that way. So have to put it in devon-modules , before spring-boot dependencies to override the jackson inside it

hohwille: Thanks for testing. Have you tried playing with the order of the BOM imports in our devon4j-bom? Maybe we can convince maven to use the desired version this way...
Do you know what `pr1` stands for? pre-release? Does not sound like a reliable version...

vapadwal: I have removed the the pr1 version, and updated with 2.9.9.20190727.

Also I have tried changing the orders in devon4j-bom but with no success

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -32,6 +32,14 @@
 
   <dependencyManagement>
     <dependencies>
+    <!-- overriding BOM of jackson -->
+    <dependency>
```
hohwille: indendation is not correct ;)

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -32,6 +32,14 @@
...
+        <groupId>com.fasterxml.jackson</groupId>
+        <artifactId>jackson-bom</artifactId>
+        <version>2.10.0.pr1</version>
```
hohwille: why are we using a different version here? Why are we not using a variable here? See also #118.

vapadwal:
> @vapadwal 👍 for this PR. As travis already verified the build this is looking very good.
> I have added some very tiny review suggestions. Could you use `${jackson.version}` also for archetype and set the variable to the same version?



> @vapadwal 👍 for this PR. As travis already verified the build this is looking very good.
> I have added some very tiny review suggestions. Could you use `${jackson.version}` also for archetype and set the variable to the same version?

Done

hohwille:
@vapadwal perfekt now 👍 
Thanks that you also solved #118 for jackson - awesome!

PR #132, opened 2019-08-05
Upgraded JUnit to 5.3.2

AbhayChandel:


maybeec:
@hohwille please see #136, which contains our upgrade experiences from my project.

hohwille:
> please see #136, which contains our upgrade experiences from my project.

I did but could not find anything regarding JUnit at all. Do you mean something else? Like some meta-information that we are updating to much and projects cannot cope? Or what am I missing?

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -79,6 +84,7 @@
 
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.22.0</version>
         <configuration>
           <excludedGroups>${devonfw.test.excluded.groups}</excludedGroups>
```
hohwille: The categories have changed to tags with different IDs. I am missing the diff to update these:
https://github.com/AbhayChandel/devon4j/blob/issue-130-update-JUnit-to-5.x/templates/server/src/main/resources/archetype-resources/pom.xml#L18

Does that still work with `excludedGroups` attributes after switching to tags?

hohwille:
Replaced by PR #151

Issue #47, opened 2018-12-11
RevisionMetadata 


hohwille:
RevisionMetadata (https://github.com/devonfw/devon4j/blob/c8f25cb073909db78d0329eda7ce44d93c9e9bae/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/RevisionMetadata.java) is located in `dataaccess` layer.
However, this is the API that is also suitable to be exposed to logic layer and even via service APIs.
Therefore this interface and most probably its type implementation (https://github.com/devonfw/devon4j/blob/9b8f9461997bb5d8048a9fec42bec3625652eec3/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/RevisionMetadataType.java) would need to be moved to `common` instead.

In order not to break compatibility, we would create a deprecated interface in the old place extending from the moved interface (and same for the type). The lazy implementation would still also implement the deprecated interface.
Then projects have the chance to migrate their code (what should also be supported by devcon migration).
In a later release we can then remove the deprecated code.

hohwille:
See current workarounds:
* https://github.com/devonfw/sonar-devon-plugin/blob/988ef16e32cc3833ee2cb773f9394edae927b302/src/main/java/com/devonfw/ide/sonarqube/common/impl/check/DevonArchitectureLayerService2DataaccessCheck.java
* https://github.com/devonfw/sonar-devon-plugin/blob/e1a825562684164cca803398fe3e3f1f28cf1bfb/src/main/java/com/devonfw/ide/sonarqube/common/impl/check/DevonArchitectureLayerLogic2Dataaccess4ComponentCheck.java

Rohit90700:
Hi @hohwille , I have create PR for this issue - https://github.com/devonfw/devon4j/pull/134 .
Note :- There is a entity AdvancedRevisionEntity.java and AdvancedRevisionListener.java copied to common because RevisionMetadataType.java having its reference.

hohwille:
Done with PR #137.
Cleanup with #138 for next release.

PR #134, opened 2019-08-08
Moved RevisionMetadata interface and its implementation to common

Rohit90700:
Issue : - https://github.com/devonfw/devon4j/issues/47 .
AdvancedRevisionEntity.java ,AdvancedRevisionListener.java also copied to devon4j-basic common because of dependency.

### DIFF
--- a/modules/basic/src/main/java/com/devonfw/module/basic/common/api/AdvancedRevisionEntity.java
+++ b/modules/basic/src/main/java/com/devonfw/module/basic/common/api/AdvancedRevisionEntity.java
@@ -1,0 +1,124 @@
+package com.devonfw.module.basic.common.api;
+
+import java.util.Date;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+import org.hibernate.envers.RevisionEntity;
+import org.hibernate.envers.RevisionNumber;
+import org.hibernate.envers.RevisionTimestamp;
+
+import com.devonfw.module.basic.common.api.entity.PersistenceEntity;
+
+/**
+ * This is a custom {@link org.hibernate.envers.DefaultRevisionEntity revision entity} also containing the actual user.
+ *
+ * @see org.hibernate.envers.DefaultRevisionEntity
+ */
+@Entity
+@RevisionEntity(AdvancedRevisionListener.class)
+@Table(name = "RevInfo")
+public class AdvancedRevisionEntity implements PersistenceEntity<Long> {
+
+  /** UID for serialization. */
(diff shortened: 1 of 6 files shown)


modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/RevisionMetadata.java
```diff
@@ -2,37 +2,14 @@
...
-   */
-  Date getDate();
-
-  /**
-   * This method gets the identifier (login) of the author who created this revision.
-   *
-   * @return the author. May be {@code null} (if committed outside user scope).
-   */
-  String getAuthor();
+@Deprecated
```
hohwille: In such case always also add JavaDoc deprecation with hint:
```
@deprecated use {@link com.devonfw.module.basic.common.api.RevisionMetadata} instead.
```

modules/basic/src/main/java/com/devonfw/module/basic/common/api/AdvancedRevisionEntity.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * This is a custom {@link org.hibernate.envers.DefaultRevisionEntity revision entity} also containing the actual user.
+ *
+ * @see org.hibernate.envers.DefaultRevisionEntity
+ */
+@Entity
+@RevisionEntity(AdvancedRevisionListener.class)
+@Table(name = "RevInfo")
+public class AdvancedRevisionEntity implements PersistenceEntity<Long> {
```
hohwille: Can you give a rationale why we had to move the Entity classes to common?
This looks odd to me. We had this once in OASP4J long time ago and got lots of user complaints from the fact that they were using other Databases and did not have control over `@GeneratedValue`, etc. or the annotation style (private field vs. getter annotation). Also devon4j with oasp4j has a long track record of private field vs. getter annotation for JPA. Chaning this would require documentation updates and some discussion.

My suggestion would be to only add the `RevisionMetadata` with its `RevisionMetadataType` implementation here and leave everything else as is.
IMHO you also need to check compatibilitiy / deprecation for existing `RevisionMetadataType` and `LazyRevisionMetadata`. Whenever we do some deprecation like this, it has to be consistent and ensure that:
* all existing types remain in the same place and get `@deprecated`
* for everything we introduce a replacement as a copy that is not coupled to the `deprecated` original. E.g. the existing `LazyRevisionMetadata` still has to implement the old `deprecated` interface as otherwise we would break compatibilitiy. This way it is therefore forever coupled with the deprecated world. We need to introduce a new implementation that only implements the new interface without being connected with the `deprecated` code.

hohwille:
Replaced by PR #137

PR #136, opened 2019-08-12
WIP: Create guide-migration-oasp3-to-devon3.1.asciidoc

maybeec:
Shared experiences from an upgrade from oasp4j 3.0.0 to devonfw 3.1.0


hohwille:
@maybeec 👍 for this valuable input.
I have some open questions:
* should we create OASP to Devonfw migration documentation in OASP documentation keeping devonfw free from OASP legacy or should we add this here?
* if we add it here, should we instead create an upgrade guide that is more like a release notes documentation and creates the upgrade from each version to the next one? What you are proposing here is the migration from oasp4j 3.0.0 to devon4j 3.1.0. What about oasp4j 3.0.0 to devon4j 3.0.0, or devon4j 3.0.0 to devon4j 3.1.0?
* almost 90% of the explained changes explained here are fully automated with `devon-ide` migration feature. IMHO the first thing to point out in such guide is this migration feature as automation is key and saves a lot of manual effort. Do we still need to manually write and explain all the steps at all? If we think that we need this, should we generate it from `devon-ide` so we have a single source of information?

maybeec:
>should we create OASP to Devonfw migration documentation in OASP documentation keeping devonfw free from OASP legacy or should we add this here?

I would put it here. You always search for the target and see what the target technology is providing for upgrade from another technology.

> If we add it here, should we instead create an upgrade guide that is more like a release notes documentation and creates the upgrade from each version to the next one? 

yes agree. I was not yet able to separate it as it was mainly written by a colleague, but actually could be quite easily resconstructed what is oasp to devonfw and devonfw to devonfw.

> almost 90% of the explained changes explained here are fully automated with devon-ide migration feature. 

partially disagree ;) We used the latest version of devcon to migrate our two applications. Anyhow, we have a master parent pom, which is declaring the dependencies to devonfw and our applications are inheriting this master parent pom only. Unfortunately, the migration tool is expecting the normal parent pom of each application to be the one generated by the archetype. 
Therefore, we had to do some steps manually, Others were not even executed as some shared libraries were even not touched (understandable). But these libraries had to be upgraded as well regarding their dependencies etc. Thus, I think it might be worth to document such steps in addition to the migration tool.

hohwille:
> partially disagree ;) We used the latest version of devcon to migrate our two applications. Anyhow, we have a master parent pom, which is declaring the dependencies to devonfw and our applications are inheriting this master parent pom only. Unfortunately, the migration tool is expecting the normal parent pom of each application to be the one generated by the archetype.
Therefore, we had to do some steps manually, Others were not even executed as some shared libraries were even not touched (understandable). But these libraries had to be upgraded as well regarding their dependencies etc. Thus, I think it might be worth to document such steps in addition to the migration tool.

My cents on this:
* First of all `devcon` is dead. The new migrator is part of `devon-ide`.
* We IMHO need to make this the most important message where to find this migrator and how to use it properly.
* You can even provide the version to start with in case your project is not following 100% devon standard has e.g. has no `devon.version` property in top-level pom - see [java commandlet doc](https://github.com/devonfw/devon-ide/blob/master/documentation/java.asciidoc)
* If things are not working properly, we need to collect this feedback and improve the migrator tool where reasonable. Sure there are some complex cases where a smart human can do the fix easily and an automated solution would be too complex for cost/usage balance.
* See also this [feature request](https://github.com/devonfw/devon-ide/issues/226)

So devonfw has a large and strong community. I think we can make this much better. Automation is key and for many things even easier than writing upgrading documentation (that should also be understood by DAUs)...

hohwille:
So IMHO we should create a general changelog similar to what we have created here:
https://github.com/devonfw/sonar-devon4j-plugin/blob/master/CHANGELOG.adoc
Input from this PR shall be used as valuable input.
Also the CHANGELOG shall point to the migration tool.

Finally we need to consider the "monolithic vs. product specific" discussion for documentation and release-notes in devonfw before I introduce more CHANGELOG files like here in devon4j.

hohwille:
I have created issue #181 for this. We will later close this PR but the provided document should be reused to write the initial CHANGELOG.

hohwille:
As we got kind of stuck here, I would suggest to merge this as is and add a link from our new changelog to the new asciidoc guide. Then I will close #181.

hohwille:
Looking at the guide again I still think it is kind of missleading. Nobody should be pointed to devcon or be told how to manually replace `oasp4j-*` dependencies to `devon4j-*` dependencies if all can be automated. WDYT?

maybeec:
I think it's a more general how to, which focuses on more complex setups where you might have some additional cross cutting libraries, which depend on oasp4j and thus need to also be migrated. Therefore, we decided to document this as an additional step in oder to get some more light into the magic of devcon or Devon ide, which will not automatically handle all cases automatically.

hohwille:
OK. We have this pending for way to long already. If we wait even longer, it will not cause any use as there are no projects left with `oasp4j`.
So I will merge this as is (even if I see that parts of the document are outdated or partially incorrect and nobody was there yet to improve the document). Projects can use this if they find it and we can still later improve the document if we find time or see the need for that.

PR #137, opened 2019-08-18
Revision meta data

vapadwal:
Fix for Issue : - #47 .
This is new PR created against PR https://github.com/devonfw/devon4j/pull/134

### DIFF
--- a/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/JpaHelper.java
+++ b/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/JpaHelper.java
@@ -2,6 +2,7 @@ package com.devonfw.module.jpa.dataaccess.api;
 
 import java.util.Collection;
 
+import com.devonfw.module.basic.common.api.RevisionMetadataType;
 import com.devonfw.module.basic.common.api.entity.GenericEntity;
 import com.devonfw.module.basic.common.api.reference.GenericIdRef;
 import com.devonfw.module.basic.common.api.reference.Ref;
@@ -72,6 +73,22 @@ public class JpaHelper {
       }
       output.add(entity);
     }
+  }
+
+  /**
+   * This is the newer implementation of RevisionMetadataType should be used with
+   * {@link com.devonfw.module.basic.common.api.RevisionMetadata}
+   *
+   * @param revEntity die {@link AdvancedRevisionEntity}.
+   * @return die Instanz von {@link RevisionMetadataType} bzw. {@code null} falls {@code revision} den Wert {@code null}
+   *         hat.
+   */
+  public static RevisionMetadataType of(AdvancedRevisionEntity revEntity) {
+
+    if (revEntity == null) {
+      return null;
+    }
+    return new RevisionMetadataType(revEntity.getId(), revEntity.getDate(), revEntity.getUserLogin());
   }
 
 }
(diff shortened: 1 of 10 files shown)


vapadwal:
> @vapadwal Thanks for taking over this PR and improving 👍
> This is now nice rearding the API. Can you generate the JavaDoc and check if it is sane? AFAIK JavaDoc taglets are case sensitive and @deprecated in JavaDoc has to be lower-case.
> 
> However, it is still not working as planned yet.
> See the APIs and implementation here:
> 
> * https://github.com/devonfw/devon4j/blob/develop/modules/jpa-envers/src/main/java/com/devonfw/module/jpa/dataaccess/base/AbstractGenericRevisionedDao.java#L13
> * https://github.com/devonfw/devon4j/blob/develop/modules/jpa-envers/src/main/java/com/devonfw/module/jpa/dataaccess/api/GenericRevisionedDao.java#L45
> * https://github.com/devonfw/devon4j/blob/develop/modules/jpa-spring-data/src/main/java/com/devonfw/module/jpa/dataaccess/api/data/GenericRevisionedRepository.java#L6
> * https://github.com/devonfw/devon4j/blob/develop/modules/jpa-spring-data/src/main/java/com/devonfw/module/jpa/dataaccess/impl/data/GenericRevisionedRepositoryImpl.java#L20
> 
> It is still producing the old depreated code which is unrealted to the new one. Therefore the new API can not yet be used.
> 
> My proposal would be:
> Let the old and deprecated code extend the new API such that the old deprected interface and class are more or less empty and just inherit from the new API types.
> For compatibility we could go these steps:
> 
> 1. keep the DAO/Repositories untouched for the moment (using deprecated code).
> 2. push out a release like this
> 3. let projects update to that release
> 4. now projects can update their code such that they use the new API and remove references to the deprecated classes from their code.
> 5. For the next devon4j release we then remove the deprecated classes and only use the new APIs.
> 6. Projects that did step 4. are now safe to upgrade to 5.
> 
> An alternative would be to simplify and go this way:
> 
> * we give up compatibility and remove the old code (instead of deprecating it)
> * we directly refactor DAO/Repository to the new API
> * we provide support in devon-ide for updating projects in our migrator
> 
> I would vote that we have migration support for our projects regarding this change anyhow. So maybe the latter approach would be much simpler. I am not entirely sure how our projects deal with breaking changes and if they are all aware of devon-ide and its migration feature.

I was able to generated javadoc with it.
I have selected first approach as it will get time for the projects to deal with the breaking changes as in 3.2.0 release they will also be new to devon-ide

modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/JpaHelper.java
```diff
@@ -74,4 +75,20 @@
...
 
+  /**
+   * This is the newer implementation of RevisionMetadataType should be used with
+   * {@link com.devonfw.module.basic.common.api.RevisionMetadata}
+   *
+   * @param revEntity die {@link AdvancedRevisionEntity}.
+   * @return die Instanz von {@link RevisionMetadataType} bzw. {@code null} falls {@code revision} den Wert {@code null}
+   *         hat.
+   */
+  public static RevisionMetadataType of(AdvancedRevisionEntity revEntity) {
```
hohwille: The static `of` method is typically used as a factory method of the same class defining the static method (as a short form for `valueOf` what is used inside the JDK as convention/pattern).
In this case `JpaHelper.of(entity)` the API is IMHO very confusing. You need to move the method to `RevisionMetadataType`. In case it should remain in `JpaHelper` you should rename this `of` method to `asRevisionMetadata` (like e.g. `asEntity`).

hohwille:
Just to be aligned. I understand the latest state of this PR such that:
* We keep the existing classes deprecated in `3.2.0` as a kind of pointer and documentation for users to simplify manually updating.
* We still need to provide a migration in devon-ide: https://github.com/devonfw/devon-ide/issues/237
* We will remove the deprecated classes in `3.3.0`.

PR #151, opened 2019-09-17
upgrade to junit5

vapadwal:
As discussed this PR is against https://github.com/devonfw/devon4j/pull/132
Have implement review comments in PR https://github.com/devonfw/devon4j/pull/132


templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -296,27 +297,15 @@
   <profiles>
     <profile>
       <id>moduletest</id>
-      <properties>
-        <devonfw.test.excluded.groups>com.devonfw.module.test.common.api.category.CategoryComponentTest,com.devonfw.module.test.common.api.category.CategorySubsystemTest,com.devonfw.module.test.common.api.category.CategorySystemTest</devonfw.test.excluded.groups>
```
hohwille: To me it seems that you removed the entire feature of enabling or disabling sets of JUnits via categories/tags.
Now these Profiles (`-Pmoduletest`) do not seem to have any effect.
What is the new way to trigger selection of JUnits via Tags in JUnit 5?
Can you please do a web-research how to configure this for JUnit 5 and surefire and come up with a new solution for our test stage profiles?

modules/test/src/main/java/com/devonfw/module/test/common/base/SystemTest.java
```diff
@@ -11,7 +10,7 @@
  *
  * @see CategorySystemTest
  */
-@Category(CategorySystemTest.class)
+@Tag("system")
```
hohwille: If we have all these tags redundatly, shouldnt we create a constant class/interface (just like `SpringProfileConstants`)?

templates/server/src/main/resources/archetype-resources/pom.xml
vapadwal: done

modules/test/src/main/java/com/devonfw/module/test/common/base/SystemTest.java
vapadwal: Created a meta annotation for all Tags

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -83,6 +84,9 @@
 
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <excludedGroups>${devonfw.test.excluded.groups}</excludedGroups>
```
hohwille: nice. So we also keep downward compatibility. Just to double check: Did you test this manually if it is working (e.g. to disable all but module tests with `-Pmoduletest`)?

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/ComponentDbTest.java
```diff
@@ -4,21 +4,21 @@
...
+import com.devonfw.module.test.common.api.category.TagComponentTest;
 import com.devonfw.module.test.common.base.clean.TestCleaner;
 
 /**
  * Combination of {@link DbTest} with {@link ComponentTest}.
  */
 @ExtendWith(SpringExtension.class)
 @TestExecutionListeners({ TransactionalTestExecutionListener.class, DependencyInjectionTestExecutionListener.class })
-@Tag("component")
+@TagComponentTest
```
hohwille: Nice approach, did not think of this. Simple string constants might have been easier but this is more sophisticated. Still people who do not want to couple to much with our dependencies can still add stuff `@Tag("component")` manually to their JUnits what was not possible with JUnit 4.

vapadwal:
> @vapadwal 👍 thank you and well done. IMHO we are ready to merge now.
> Could you finally confirm that you tested the profiles and opt-in/-out of tests via tags?

Tested with profiles
mvn clean install -P systemtest
mvn clean install -P componenttest
 mvn clean install -P moduletest
 mvn clean install -P subsystemtest

PR #155, opened 2019-09-19
Update master-devon4j.asciidoc

jambulud:


hohwille:
I already asked this (on Yammer I think): Can someone explain me what this `master-devon4j.asciidoc` is all about?
The "master" that we have defined already exists here:
https://github.com/devonfw/devon4j/blob/develop/documentation/devon4j-doc.asciidoc

Why do we need another "master" document?
Seems redundant to me... What am I missing?

Jorge-Dacal:
Hi Jörg,

To generate the devonfw guide we needed a document in each repository with the content for the guide.

The problem is that in each repository the filename was different and using different patterns.

I think that in the end it was decided to put an asciidoc in each repository called master-$repoName.asciidoc

hohwille:
Thanks for the feedback. I will discuss this with santos and others on next occasion so we might be able to avoid redundancies and maintenance issues. To not furhter block anything I will merge...

PR #76, opened 2019-04-18
Adding security-jwt module

ssarmokadam:
Hi @hohwille ,

This PR is for adding security-jwt module as per our discussion. I have implemented it for RSA algo by default and configuring properties in application.properties file.

PR #157, opened 2019-09-22
Adding jwt module

vapadwal:
As discussed this is backup of PR #76 
with latest updates

Please give your review comments here


modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -0,0 +1,140 @@
...
+					LOG.info("Token is expired");
+				}
+			}
+
+
+			String user = tokenClaims.get("sub").toString();
+			// If token do not throw any exception above it is valid and we need to create
+			// authentication object from it
+
+			authentication = new UsernamePasswordAuthenticationToken(user,
```
hohwille: So this code is providing a `UsernamePasswordAuthenticationToken`. This does not really make sense for me as there is no password. Also the implementation of the authentication manager seems to require credentials.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/config/KeyStoreAccessImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+	@Override
+	public PublicKey getPublicKey() {
+		PublicKey publicKey = null;
+		try {
+			Certificate certificate = keyStore.getCertificate(keyStoreProps.getKeyAlias());
+			publicKey = certificate.getPublicKey();
+		} catch (KeyStoreException e) {
+
+			LOG.error(e.getMessage());
```
hohwille: Stacktraces are always important and valuable information for analysis and debugging. This code is dumping this information and therefore not production ready.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/config/KeyStoreAccessImpl.java
```diff
@@ -0,0 +1,94 @@
...
+			Key key = keyStore.getKey(alias, "changeit".toCharArray());
+			 if (key instanceof PrivateKey) {
+				 return (PrivateKey) key;
+			    }
+		} catch (UnrecoverableKeyException e) {
+			e.printStackTrace();
+		} catch (KeyStoreException e) {
+			e.printStackTrace();
+		} catch (NoSuchAlgorithmException e) {
+			e.printStackTrace();
```
hohwille: never do that.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -0,0 +1,140 @@
...
+		case "Elliptic Curve":
+			return null;
+
+		case "HMAC":
+			return null;
+
+		case "RSA":
+			return new RsaVerifier((RSAPublicKey) publicKey);
+
+		default:
```
hohwille: This seems pretty much incomplete for our `SignatureAlgorithm` enum. Also we should consider making the Enum the factory for `SignatureVerifier` so if you add a new enum constant you are forced to also provide this. Of course this will bundle the Enum with more implementation details and therefore do tight coupling but otherwise we have the same issue here but with the disadvantage the we easily miss a new option here.

modules/security-jwt/src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
+#security.authentication.keystore.keystorelocation=C:/Devon-dist_3.0.0/jwtkeystore.p12
```
hohwille: we should make explicit that these are only some examples. Further, it only makes sense to add reasonable defaults here. Some more reasonable default could be `~/keystore.p12` or `~/jwt-keystore.p12` as long as we would support resolving `~` to `user.home` system property value (do not know if Java NIO already does that or we still would have to code that explicitly).

modules/security-jwt/src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
...
+#security.authentication.keystore.keystoretype=PKCS12
+#security.authentication.keystore.password=123456
```
hohwille: For passwords there is never a reasonable default. This should be a secret that differes per environment. So this config should only go to `config/application.properties` but shall be removed here.

hohwille: https://github.com/devonfw/devon4j/blob/develop/documentation/guide-configuration.asciidoc#externalized-configuration

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/config/JwtAuthenticationProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+	public void setUserDetailsService(UserDetailsService userDetailsService) {
+		this.userDetailsService = userDetailsService;
+	}
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		UsernamePasswordAuthenticationToken auth = (UsernamePasswordAuthenticationToken) authentication;
+		UserDetails userDetails = this.userDetailsService.loadUserByUsername(auth.getName());
+		UsernamePasswordAuthenticationToken output = new UsernamePasswordAuthenticationToken(userDetails,
```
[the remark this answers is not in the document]
vapadwal: I am analyzing this class, actually do we required this class, which is implements AuthenticationProvider. Actually I feel provider should left with the user who is using our jwt module

vapadwal:
@hohwille  please find the updated PR for Jwt, I have tested this PR against my-thai-star.

The configuration required in my-thai-star will be as follows 

- Add dependency in mtsj-core pom
`<dependency>
        <groupId>com.devonfw.java.modules</groupId>
        <artifactId>devon4j-security-jwt</artifactId>
        <version>3.2.0-SNAPSHOT</version>
    </dependency>`
    

- in BaseWebSecurityConfig.java
  Add below 2 methods 
`  @Bean
  public JwtLoginFilter jwtLoginFilter() throws Exception {

    JwtLoginFilter jwtLoginFilter = new JwtLoginFilter("/login", authenticationManager(), this.userDetailsService);
     return jwtLoginFilter;
  }

  @Bean
  public JwtAuthenticationFilter jwtAuthenticationFilter() throws Exception {

    JwtAuthenticationFilter jwtAuthenticationFilter = new JwtAuthenticationFilter();
    return jwtAuthenticationFilter;
  }

- In configure method of `BaseWebSecurityConfig
  add the above 2 filters
`.addFilterBefore(jwtLoginFilter(), UsernamePasswordAuthenticationFilter.class)`
`addFilterBefore(jwtAuthenticationFilter(), UsernamePasswordAuthenticationFilter.class);`

-  Add @ComponentScan in Springbootapp for the above moduel

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/JwtSignatureAlgorithm.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private final String description;
+
+  private final String familyName;
+
+  private final String jcaName;
+
+  private final boolean jdkStandard;
+
+  private JwtSignatureAlgorithm(String value, String description, String familyName, String jcaName,
```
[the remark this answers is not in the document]
vapadwal: This enum can be usefull - 
Basically in validateTokenAndSignature
JwtHelper.headers(token).get("alg") - this returns the algorithm from the header - RS256 // see https://en.wikipedia.org/wiki/JSON_Web_Token

so first argument value is for that. and other args are the mapping for value, agreed that description is not needed

so we have currently 3 SignatureVerif (EllipticCurveVerifier, MacSigner , RsaVerifier)

The enum basically maps this 
HS256,HS384,HS512 maps to HMAC(MacSigner )
RS256,RS384, RS512,PS256,,PS384,PS512 maps to RSA(RsaVerifier)
ES256,ES384,ES512 maps to EllipticCurve(EllipticCurveVerifier)

So signatureVerifierFactory would be something like this - 
` private SignatureVerifier signatureVerifierFactory(PublicKey publicKey, String algorithm) {

    SignatureVerifierAlgorithm signatureVerifierAlgorithm = SignatureVerifierAlgorithm.valueOf(algorithm);
    switch (signatureVerifierAlgorithm) {
      case ELLIPTICCURVE:
        return new EllipticCurveVerifier((ECPublicKey) publicKey, algorithm);
      case HMAC:
        return new MacSigner(publicKey.getFormat());// Need to check if it return the key
      case RSA:
        return new RsaVerifier((RSAPublicKey) publicKey, algorithm);

      default:
        return new RsaVerifier((RSAPublicKey) publicKey, algorithm);
    }`

So as u suggested we can completely remove this enum and configure this two 
"RSA", "SHA256withRSA" as properties.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/SignatureVerifierAlgorithm.java
```diff
@@ -0,0 +1,37 @@
+package com.devonfw.module.security.jwt.util;
+
+/**
+ * Enum defining available Algorithms for JWT comparison.
+ *
+ * @since 3.2.0
+ */
+public enum SignatureVerifierAlgorithm {
```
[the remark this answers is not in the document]
vapadwal: agreed

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenConfigProperties.java
```diff
@@ -0,0 +1,231 @@
...
+   */
+  public void setIssuer(String issuer) {
+
+    this.issuer = issuer;
+  }
+
+  /**
+   * @return expiratioHours
+   */
+  public Integer getExpiratioHours() {
```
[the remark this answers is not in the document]
vapadwal: mistake copy paste

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -1,140 +1,104 @@
...
+   */
+  public Authentication validateTokenAndSignature(String token) {
 
-		String claims = jwt.getClaims();
-		try {
-			tokenClaims = this.objectMapper.readValue(claims, Map.class);
+    Authentication authentication = null;
 
-			LOG.info("Token Claims " + tokenClaims.toString());
+    com.devonfw.module.security.jwt.sign.JwtSignatureAlgorithm jwtSignatureAlgorithm = this.jwtSignatureAlgorithmFactory
```
hohwille: user import for `JwtSignatureAlgorithm`.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -1,140 +1,104 @@
...
+    String claims = jwt.getClaims();
+    try {
+      this.tokenClaims = this.objectMapper.readValue(claims, Map.class);
 
-			}
+      LOG.info("Token Claims " + this.tokenClaims.toString());
 
-			if(tokenClaims.get("exp") !=null) {
-				Instant expiration_time = Instant.ofEpochSecond(((Number) tokenClaims.get("exp")).longValue());
+      Instant now = Instant.now();
```
hohwille: here `java.time` is used instead of `java.util.Date` - nice 👍

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -1,140 +1,104 @@
...
-				if (now.isAfter(expiration_time)) {
-					// Throw exception
-					LOG.info("Token is expired");
-				}
-			}
+      Object nbfObj = this.tokenClaims.get("nbf");
+      Object expObj = this.tokenClaims.get("exp");
 
+      if (nbfObj != null) {
+        Instant nbf = Instant.ofEpochSecond(((Number) nbfObj).longValue());
```
hohwille: use a private function to get the claim value by key as `Long`?

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -1,140 +1,104 @@
...
-				Instant expiration_time = Instant.ofEpochSecond(((Number) tokenClaims.get("exp")).longValue());
+      Instant now = Instant.now();
 
-				if (now.isAfter(expiration_time)) {
-					// Throw exception
-					LOG.info("Token is expired");
-				}
-			}
+      Object nbfObj = this.tokenClaims.get("nbf");
+      Object expObj = this.tokenClaims.get("exp");
```
hohwille: So here you are using hardcoded string literals for the claims? Why can they be configured then?
As I stated above define constants for that. Maybe the JWT API already provides such constants or even methods for the standardized claims.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenExtractor.java
```diff
@@ -1,140 +1,104 @@
...
-		case "HMAC":
-			return null;
+      List<GrantedAuthority> authorities = new ArrayList<>();
+      for (String role : roles) {
+        authorities.add(new SimpleGrantedAuthority(role));
+      }
 
-		case "RSA":
-			return new RsaVerifier((RSAPublicKey) publicKey);
+      authentication = new UsernamePasswordAuthenticationToken(user, null, authorities);
```
hohwille: The fabrication of the Authentication should go to a replacable component or at least a protected method that can be overridden. Projects will often require custom code here and might have their own custom implementation of Authentication that needs to be used.

modules/security-jwt/src/main/resources/application.properties
```diff
@@ -1,3 +1,20 @@
-#security.authentication.keystore.keystorelocation=C:/Devon-dist_3.0.0/jwtkeystore.p12
+## below properties need to be configured in config\application.properties
```
hohwille: do not provide `application.properties` as part of this module at all.
Add this to the application template and/or documentation instead.

modules/security-keystore/pom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>com.devonfw.java.dev</groupId>
+		<artifactId>devon4j-modules</artifactId>
+		<version>dev-SNAPSHOT</version>
+	</parent>
```
hohwille: devonfw indendation size is 2 and not 4. Are you using `devonfw-ide`?

modules/security-keystore/src/main/java/com/devonfw/module/security/keystore/config/KeyStoreAccessImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return keyStore;
+  }
+
+  @Override
+  public PublicKey getPublicKey() {
+
+    PublicKey publicKey = null;
+    try {
+      Certificate certificate = getKeyStore().getCertificate(this.keyStoreConfigProperties.getKeyAlias());
```
hohwille: as stated elsewhere this information has to come from outside. The JWT config should configure the alias for JWT. Other modules that might want to use `KeyStoreAccess` in the future will use a different alias.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/sign/RsaJwtSignatureAlgorithmImpl.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public String getName() {
+
+    return ALGORITHM;
+  }
+
+  @Override
+  public Signer createSigner() {
+
+    return new RsaSigner((RSAPrivateKey) this.keyStoreAccess.getPrivateKey());
```
hohwille: We need to be careful. When is this code used? Most apps will not issue JWTs and do not have a private key. Therefore such configuration will not be present and this method would currently end up in a runtime exception.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/util/TokenConfigProperties.java
```diff
@@ -0,0 +1,249 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * This class contain properties related to {@link TokenAuthenticationUtil}
+ *
+ * @since 3.2.0
+ */
+@Configuration
+@ConfigurationProperties(prefix = "security.authentication.token")
+public class TokenConfigProperties {
```
hohwille: This class does not define any defaults. So we should follow the `convention-over-configuration` approach and define defaults where suitable (e.g. expiration, alias for private/public key, could have reasonable defaults. Some other properties might not have defaults such as `issuer` - or should we use `devonfw` as default?).

modules/security-keystore/src/main/resources/application.properties
```diff
@@ -0,0 +1,20 @@
+## below properties need to be configured in config\application.properties
```
hohwille: Same here. Delete this file and provide as doc/template instead.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/sign/RsaJwtSignatureAlgorithmImpl.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Override
+  public SignatureVerifier createVerifier() {
+
+    return new RsaVerifier((RSAPublicKey) this.keyStoreAccess.getPublicKey());
```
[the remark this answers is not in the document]
vapadwal: I am not sure if i understand it correctly, currently for getting public key i have use api keystore.getCertificate(alias). And for getting private key I have used keystore.getKey(alias,password)

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/authentication/DefaultAuthenticationTokenDetailsImpl.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+
+/**
+ * Default Implementation of {@link AuthenticationTokenDetails}
+ *
+ * @since 3.3.0
+ */
+@Named
+public class DefaultAuthenticationTokenDetailsImpl implements AuthenticationTokenDetails {
```
hohwille: This class is used as a singleton spring bean but is stateful and therefore not thread-safe. This can not work.

vapadwal:
As discussed added  - io.jsonwebtoken - jjwt and refactored some code as needed. Now devonfw-jwt is using jjwt api

hohwille:
@vapadwal thanks for your improvements. 
We still need to merge from `jwtbr` to `develop` so the latest polishing can happen there.

Issue #96, opened 2019-06-03
Security updates for 3.2.0


hohwille:
I just created a new devon4j `3.0.1` project and run the depdendency check what gave me the following errors:
```
javax.el-api-3.0.0.jar (cpe:/a:oracle:glassfish:3.0.0, javax.el:javax.el-api:3.0.0) : CVE-2013-2566
javax.el-2.2.6.jar (cpe:/a:oracle:glassfish:2.2.6, org.glassfish.web:javax.el:2.2.6) : CVE-2013-2566
spring-security-core-5.0.7.RELEASE.jar (cpe:/a:pivotal_software:spring_security:5.0.7, org.springframework.security:spring-security-core:5.0.7.RELEASE) : CVE-2019-3795
jackson-databind-2.9.6.jar (com.fasterxml.jackson.core:jackson-databind:2.9.6, cpe:/a:fasterxml:jackson-databind:2.9.6) : CVE-2019-12086
commons-beanutils-1.9.1.jar (commons-beanutils:commons-beanutils:1.9.1, cpe:/a:apache:commons_beanutils:1.9.1) : CVE-2014-0114
devon4j-cxf-client-3.0.1.jar (cpe:/a:apache:cxf:3.0.1, com.devonfw.java.modules:devon4j-cxf-client:3.0.1) : CVE-2014-3623
tomcat-embed-core-8.5.32.jar (cpe:/a:apache:tomcat:8.5.32, org.apache.tomcat.embed:tomcat-embed-core:8.5.32) : CVE-2019-0199, CVE-2019-0221, CVE-2019-0232
```

We should check for both `3.0.x` as well as `3.1.x` releases that we fix what we can fix easily from this list.

hohwille:
We need to check what we can fix easily without any harm or risk.
Just hunting for CVEs can also lead to lowered quality if you do not care what updates you are introducing. Further, some CVEs just list vulnerability without a fixed version of the library available nor any real attac vector in regular usage of the library (but only exotic scenarios)...

hohwille:
Some (most?) issues should already be resolved for `3.1.0-SNAPSHOT`.

hohwille:
Kind of duplicate of issue #131

hohwille:
#131 was resolved for 3.1.1 and has been merged so it will be included also in 3.2.0.
[...]

PR #163, opened 2019-10-15
#96: update to spring-boot 2.1.9

hohwille:
update to spring-boot `2.1.9.RELEASE` for issue #96

### DIFF
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -73,7 +73,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>28.0-jre</version>
+        <version>${guava.version}</version>
       </dependency>
       <!-- Library with advanced collection support -->
       <dependency>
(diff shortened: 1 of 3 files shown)


hohwille:
We still seem to have guava 17.0 with this.
We need to override it to have guava version `27.1-jre`.
AFAIK we are getting guava from `query-dslcore`. Maybe there is also an update of querydsl available. Otherwise we should try to override guava in BOM or according module (`basic-core`).

hohwille:
Build is looking good now. Should we include this fix into the release?
According to my tests all seems to work.

hohwille:
I have rerun the OWASP check and guava seems to be fixed now. I also verified that now the archetype uses guava `28.1-jre` as intendet. So at least one CVE fixed and some other updates included.

vapadwal:
I have done the basic testing creating a project with this and it seems we dont have issue here

hohwille:
I also rely on spring-boot who did good quality assurance. Also we still have the ability to ship a bugfix release. Being agile and keep updating is after all better than always being frigthened about what could break or go wrong. As we all have tested and automated tests are also fine, I will go for it. Thanks @vapadwal for testing and review.

Issue #166, opened 2019-11-04
JUnit5 not working (in Eclipse)


hohwille:
### Expected behavior

As a developer, I want to write and run JUnit tests so that I can assure quality.

### Actual behavior

When I run my JUnits in Eclipse, I get this error:

```
java.lang.NoSuchMethodError: org.junit.platform.launcher.Launcher.execute(Lorg/junit/platform/launcher/TestPlan;[Lorg/junit/platform/launcher/TestExecutionListener;)V
```
See also here:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=545849

Seems this was never tested before releasing. Is Eclipse not used for MTS(J)?

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  Create devon4j app from 3.2.0 app template
2. Launch JUnit in Eclipse

### Related/Dependent Issues
See #151 

### Comments/Hints:
IMHO the problem is that `devon4j-test` comes with dependencies such as `junit-platform-runner` and `junit-platform-launcher` in an obsolete version `1.3.2`. These dependencies should either not be shipped with devonfw libs if that is possible or we need to ship a recent version (at least `1.4.2`)

### Affected version:
`devon4j:3.2.0`

hohwille:
due to the urgency, I am taking over this issue.

hohwille:
We are including spring-boot BOM in version `2.1.9.RELEASE` which already includes junit in a wrong/broken version:
```
    <junit-jupiter.version>5.3.2</junit-jupiter.version>
```

hohwille:
I got this working by upgrading to spring-boot `2.0.0` but now `@Inject` is not working into JUnits so `DependencyInjectionTestExecutionListener` stopped working.
Seems that JUnit5 is together with its ecosystem still in some kind of experimental state and we migrated too early?

hohwille:
Seems I got the cause why `DependencyInjectionTestExecutionListener` stopped working. `@SpringJUnitConfig` is the magic link that was missing (technically `@ContextConfiguration`):
https://docs.spring.io/spring/docs/current/spring-framework-reference/testing.html#integration-testing-annotations-junit-jupiter

hohwille:
Fixed with PR #168

PR #168, opened 2019-11-05
#166: fixes for JUnit5 to work properly in all environments

hohwille:
Fix for #166 
Please review. I did not intend to also have the impact on upgrading validation-api but could not find another way as also hibernate-validator was updated by switching the BOM order what was required to overrule spring-boot-dependencies BOM what ships with "broken" junit versions.

### DIFF
--- a/modules/test/pom.xml
+++ b/modules/test/pom.xml
@@ -10,7 +10,7 @@
   <artifactId>devon4j-test</artifactId>
   <version>${devon4j.version}</version>
   <name>${project.artifactId}</name>
-  <description>Module with code and configuration for tests of the Open Application Standard Platform for Java (devon4j).</description>
+  <description>Module with code and configuration for tests.</description>
 
   <dependencies>
     <dependency>
--- a/modules/test/pom.xml
+++ b/modules/test/pom.xml
@@ -24,6 +24,12 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
--- a/modules/test/pom.xml
+++ b/modules/test/pom.xml
@@ -15,11 +15,7 @@
   <dependencies>
     <dependency>
      <groupId>org.junit.jupiter</groupId>
-     <artifactId>junit-jupiter-engine</artifactId>
-    </dependency>
-    <dependency>
-     <groupId>org.junit.platform</groupId>
-     <artifactId>junit-platform-runner</artifactId>
+     <artifactId>junit-jupiter</artifactId>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -233,7 +233,7 @@
       <dependency>
         <groupId>org.junit</groupId>
         <artifactId>junit-bom</artifactId>
-        <version>5.5.2</version>
+        <version>${junit.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -12,7 +12,7 @@
   <version>${devon4j.version}</version>
   <packaging>pom</packaging>
   <name>${project.artifactId}</name>
-  <description>Dependencies (BOM) of the Open Application Standard Platform for Java (devon4j) based on spring boot.</description>
+  <description>Dependencies (BOM) of devon4j based on spring boot.</description>
   <url>https://devon4j.github.io/</url>
   <inceptionYear>2014</inceptionYear>
 
--- a/boms/bom/pom.xml
+++ b/boms/bom/pom.xml
@@ -183,7 +183,7 @@
       <dependency>
         <groupId>javax.validation</groupId>
         <artifactId>validation-api</artifactId>
-        <version>1.1.0.Final</version>
+        <version>2.0.1.Final</version>
       </dependency>
       <!-- JAXB -->
       <dependency>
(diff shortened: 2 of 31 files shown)


modules/test/pom.xml
```diff
@@ -28,6 +24,12 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
```
maybeec: spring boot is depending on JUnit 4 or why do you exclude it here and overwrite the junit version above?

hohwille: spring-boot-starter-test ships with JUnit4 (for compatbility). You can also mix JUnit4 tests and Junit5 tests. However both have their own `@Test` annotation in different packages what caus lots of confusion in auto-completion.
Above I am not adding `junit4` as dependency but entire BOM of junit what contains the versions of all junit4 and junit5 artifacts.

maybeec: ok got it. I thought they already migrated to junit 5

vapadwal: @hohwille I have test it though few changes will be required

1. There is a compilation error here as you see the build is failing, 

In class AccessControlSchemaTest
need to imoprt 
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotNull;

instead of 
import org.junit.Assert; 

Also required change in method testGroupTypes in same test
according to above imports

2. When we make a project with latest devon4j (with this PR), we still get the same error

What i tried is override the launcher in template devon4j\templates\server\src\main\resources\archetype-resources\core\pom.xml

Currently 1.3.2 is creating the problem

<dependency>
        <groupId>org.junit.platform</groupId>
        <artifactId>junit-platform-launcher</artifactId>
        <version>1.4.2</version>
      </dependency>

With the above changes it worked

hohwille:
@maybeec and @vapadwal thanks for your feedback. I will rework the template as well.

hohwille:
still too slow while doing the training. All tests now completely migrated to JUnit5. However now getting this error:
```
org.apache.maven.plugin.MojoFailureException: groups/excludedGroups require TestNG, JUnit48+ or JUnit 5 on project test classpath
```

I start to love this topic ;)

hohwille:
From my end now everything is fine. I will merge this tomorrow. If there is any concern or feedback please let me know asap.

vapadwal:
@hohwille I have created a devon4j project using template, and open that project using eclipse. getting the below error when i run any junit in eclipse
java.lang.NoClassDefFoundError: org/junit/platform/commons/PreconditionViolationException
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.createUnfilteredTest(JUnit5TestLoader.java:75)
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.createTest(JUnit5TestLoader.java:66)
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.loadTests(JUnit5TestLoader.java:53)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:525)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:763)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:463)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:209)
Caused by: java.lang.ClassNotFoundException: org.junit.platform.commons.PreconditionViolationException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	... 7 more

hohwille:
@vapadwal thanks for testing and feedback. I tested the same and it worked yesterday. However, I will try once more and try to reproduce...

hohwille:
I can reproduce this error now. Thanks so much for pointing this out. It is actually a huge pain what JUnit5 has done here. Tests run in maven but fail in eclispse. No way one can detect this with CI. No clue if after fixing we might have other errors in different IDEs. If we would not have pushed out a release with JUnit5 and added migrations in devonfw-ide, I would rather rollback and wait for more stability of JUnit...

hohwille:
OMG - what have they done?
Screenshot 2019-11-07 at 08 56 42

maybeec:
might it be worth to toggle junit5 out of the release? i am fearing that this might have some impact we do not know at the moment.
Then in the next dev branch enable it again and have one release time for testing.

hohwille:
At least I found the source of the problem:
Again it is a dependencyManagement failure as `junit-platform-commons` should be `1.5.2` but `1.3.2` is used.

hohwille:
Reverting this is IMHO even more complex and risky for the release. As I said `devonfw-ide` contains migrations from JUnit4 to 5, also we have release-notes, announcements, etc. Further having release 3.2.0 out with JUnit5, then shipping 3.2.1 with Junit4 and again 3.2.2 with JUnit5 will IMHO cause more harm than cure.
At least I can not copy with all these revertings and for me it is easier to fix this and go ahead.

maybeec:
ok, did not know the context at all. so then I agree.

vapadwal:
I know its kind of hack , What i tried just override the dependency in core pom and it worked

`<groupId>org.junit.platform</groupId>
     <artifactId>junit-platform-launcher</artifactId>
      <version>1.4.2</version>
       </dependency>`

hohwille:
Yep, that should do it. Thanks for the hint. However, I think it is better to add the workaround to `devon4j-test` POM.
Here my analysis:
https://repo1.maven.org/maven2/org/junit/junit-bom/5.5.2/junit-bom-5.5.2.pom
```
<dependency>
        <groupId>org.junit.jupiter</groupId>
        <artifactId>junit-jupiter</artifactId>
        <version>5.5.2</version>
</dependency>
<dependency>
        <groupId>org.junit.jupiter</groupId>
        <artifactId>junit-jupiter-api</artifactId>
        <version>5.5.2</version>
</dependency>
```

However, the dependency version gets overriden by some POM again:
Screenshot 2019-11-07 at 10 37 22

So I assume the problem is that the JUnit BOM is missing in the archetype toplevel POM.

hohwille:
Now also in the archetype JUnits work with Eclipse.

hohwille:
Manually verified:
```
devon java create org.example.apptest -DarchetypeVersion=3.2.1-SNAPSHOT
```
imported into eclipse and run `src/test/java` as Junit test in all three modules.
Only fails in server as there is no test (`src/test/java/org.example.apptest` is empty).

Issue #177, opened 2019-11-28
Switch to new maven-parent


hohwille:
As a devonfw evangelist, I want to use a central maven parent pom for devonfw so that we can reduce lots of redundancies and simplify maintainance.
See https://github.com/devonfw/maven-parent/issues/2

hohwille:
From PR #178 :

> However, we might also use the occasion to revisit the archetype parent POM and align this with our new maven parent (versions of plugins, etc.). We could even consider #118 and expose versions for all plugins in our maven-parent pom in a next release and reuse exactly these versions for archetype parent.

This IMHO is something we should do.

hohwille:
Done.

PR #178, opened 2019-11-28
#177: updated maven to new parent

hohwille:
* #177: updated maven config
* also fixed some leftovers that had been wrong from the start in oasp4j/devon4j


hohwille:
So I am not using the new parent POM in the loplevel POM of the maven archetype because:
* projects should have 100% freedom to adopt their versions and settings for maven
* projects should not inherit github specific settings such as scm, issueManagement, etc.
* projects should not inherit Apache Software License 2.0 from devonfw.

However, we might also use the occasion to revisit the archetype parent POM and align this with our new maven parent (versions of plugins, etc.). We could even consider #118 and expose versions for all plugins in our `maven-parent` pom in a next release and reuse exactly these versions for archetype parent.
WDYT?

PR #179, opened 2019-11-29
#177: updated parent #118: avoid redundancies

hohwille:
* #177 parent updated to version `2`
* #118 avoid redundant versions and inherit them from parent POM into archetype.


vapadwal:
checked with mvn install for devon4j and tried creating a project with it. everything seems to be fine

Issue #183, opened 2019-12-06
Error in executing the project created with devon4j


vapadwal:
I tried creating a project using latest devon4j. the project gets created successfully 

but when i run the root pom of the project created get the error -  

Non-resolvable import POM: Could not find artifact com.devonfw.java.boms:devon4j-minimal-bom:pom:dev-SNAPSHOT @ com.devonfw.java.boms:devon4j-bom:${devon4j.version}

hohwille:
Downgraded to flatten-maven-plugin version `1.0.1` but getting the same error.
I have another idea what could be the problem. Checking...

hohwille:
Got it:
https://github.com/devonfw/devon4j/blob/1e1d37e62c255e2be723e97f4df861be3971b892/boms/bom/pom.xml#L24
and 
https://github.com/devonfw/devon4j/blob/1e1d37e62c255e2be723e97f4df861be3971b892/boms/minimal/pom.xml#L22

hohwille:
What I do not really get is why the maven archetype did not fail as it includes integration tests.
Seems that these tests are not as realistic as I hoped...

hohwille:
Got it fixed. Will directly commit instead of creating PR as develop is broken anyhow and I better do not waste more time till it is fixed.

hohwille:
OK. Fixed. I assume all is done. Hence, closing...
@vapadwal could you double check on your machine to get 100% sure. If still failing, please reopen.

vapadwal:
Tested it on your branch https://github.com/hohwille/devon4j , and its working now

PR #184, opened 2019-12-10
#183: fixed BOMs so devon4j and archetype can be used again

hohwille:
See #183


vapadwal:
Tested its working

hohwille:
@vapadwal thanks for approval. I will wait till build is green and then squash merge it.

Issue #190, opened 2020-01-14
Rework of spring-batch integration


maihacke:
The current integration of spring-batch is not working. It has to be further adopted to changes introduced with spring boot etc.
This also affects MTS.

hohwille:
PR #191 and #199 are merged. Hence, considered done.

PR #191, opened 2020-01-15
Rework of devon4j-batch module

maihacke:
Rework how jobs are loaded from context (see https://github.com/devonfw/devon4j/issues/190). Made it more similar to SpringsCommandLineJobRunner, which also loads jobs by bean name.
Also the injection of required spring batch-beans as been aligned with
CommandLineJobRunner.

### DIFF
--- a/modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
+++ b/modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
@@ -83,7 +83,12 @@ import org.springframework.util.StringUtils;
  * </pre>
  * </p>
  * <p>
+ *
+ * <b>Deprecated</b> Since spring batch and spring boot are nicely integrated it is possible to start batches without
+ * any custom launcher. This is documented in the current devonfw batch documentation. This launcher is no longer
+ * required and will be removed in one of the next releases.
  */
+@Deprecated
 public class SpringBootBatchCommandLine {
 
   private static final Logger LOG = LoggerFactory.getLogger(SpringBootBatchCommandLine.class);
--- a/modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
+++ b/modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
@@ -29,26 +29,60 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.util.StringUtils;
 
 /**
- * Launcher for launching batch jobs from the command line when Spring Boot is used. Similar to the
- * {@link CommandLineJobRunner}, which does not work very well with Spring Boot.
- * <p>
- * Do not use this class if Spring Boot is not used!
- * <p>
- * It expects the full class name of the Spring Boot configuration class to be used as first argument, the class/XML
- * file for configuring the job as second argument and the job name as third.<br>
+ * Launcher for launching batch jobs from the command line when Spring Boot is used. It is somewhat similar to the
+ * {@link CommandLineJobRunner}. The main difference is, that this launcher disables the web-app for the spring context.
+ *
+ * It expects the full class name of the Spring Boot configuration class to be used as first argument and the jobs
+ * beanname as the second.<br>
  * Moreover parameters can be specified as further arguments (convention: key1=value1 key2=value2 ...).
  * <p>
  * Example:<br>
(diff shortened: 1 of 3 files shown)


maihacke:
New examples are available in https://github.com/devonfw/my-thai-star/pull/321

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
```diff
@@ -15,36 +18,77 @@
...
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+ * }
+ * </pre>
+ * </p>
+ * <p>
+ *
+ * <b>Deprecated</b> Since spring batch and spring boot are nicely integrated it is possible to start batches without
```
hohwille: Please use `@deprecated` for JavaDoc.

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
```diff
@@ -97,29 +141,25 @@ public static void main(String[] args) throws Exception {
...
   }
 
   private static void handleIncorrectParameters() {
 
     LOG.error("Incorrect parameters.");
     LOG.info("Usage:");
-    LOG.info("java com.devonfw.module.batch.common.base.SpringBootBatchCommandLine"
-        + " <SpringBootConfiguration> <BatchJobConfiguration>" + " <JobName> param1=value1 param2=value2 ...");
+    LOG.info(
+        "java -jar my-app-batch-bootified.jar <SpringBootConfiguration> <JobName> param1=value1 param2=value2 ...");
```
hohwille: Just a thought in this:
If we deprecated this anyhow, does it make sense to change the CLI API after all?
Maybe it is even more useless without that change. So just a note to consider or ignore.

modules/batch/src/main/java/com/devonfw/module/batch/common/impl/JobLauncherWithAdditionalRestartCapabilities.java
```diff
@@ -21,7 +22,10 @@
  * the 'run.id' parameter). It is actually just a convenience functionality so that the one starting batches does not
  * have to change the parameters manually.
  *
+ * <b>Deprecated</b> Using springs {@link RunIdIncrementer} is sufficient for most use cases.
```
hohwille: Same here: `@deprecated`.

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
maihacke: done

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
maihacke: you're right. i'll change that.

modules/batch/src/main/java/com/devonfw/module/batch/common/impl/JobLauncherWithAdditionalRestartCapabilities.java
maihacke: Done

modules/batch/src/main/java/com/devonfw/module/batch/common/base/SpringBootBatchCommandLine.java
maihacke: I rivisited this topic.
The problem is that the release implementation the launcher in devonfw will not work together with spring boot. The idea for batches in devon was always to reuse as much as possible of your application code for implementing your batches. We provided solution which accomplished that in the first release of devon4j-batch.
When introduction spring boot and java based configs this solution broke and there was never a real solution how to start your batch in the "context" of your boot app, how to handle classpath when having a bootified jar and so on.
So the current change in this PR makes things working again with the devon batch launcher and spring boot apps. Since this is not an optimal solution (the launcher is not required), I fixed and deprecated it.
We could leave the implementation as is, but we cannot document how to use it (since it is not working ;-))

hohwille:
After merge of PR #199 this now has conflicts. Can you update your feature branch, @maihacke ? Thanks.

maihacke:
As soon as we have merged this, we should rework the devon4j-template.

maihacke:
Rebased this PR to resolve conflicts, everything should be finde now.

Issue #160, opened 2019-10-14
Clarify how to proceed with master-devon4j.asciidoc vs. devon4j-doc.asciidoc


hohwille:
As a maintainer of devon4j, I want to avoid redundancies and duplications in the documentation so that maintenance efforts are low and documentation stays consistent.

From [PR 155](https://github.com/devonfw/devon4j/pull/155) we have realized and discussed that there are two main files for the documentation that more or less bundle all the other guides and chapters to one file:
1. https://github.com/devonfw/devon4j/blob/develop/documentation/devon4j-doc.asciidoc
2. https://github.com/devonfw/devon4j/blob/develop/documentation/master-devon4j.asciidoc

While 1. was created by the devon4j team and is therefore maintained by them the 2. was added externally for the devon-guide. If there is no other way I am happy to support the devon-guide and take over 2. but so far I still do not get why we need two files for more or less the same thing. Maybe our overall devonfw documentation strategy is still unclear to me. This issue is to track the clarification of this circumstance.

sjimenez77:
We created the second one to be able to generate the PDF of the devonfw guide without breaking the other PDF generation as we were researching how to do it accordingly. 

Having said that, it makes sense to have only one master document that could work for both PDFs and the website documentation automated generation.

hohwille:
So IMHO the only pending issue here is to figure out if we 
* either can delete `devon4j-doc.asciidoc` and tell docgen to generate `master-devon4j.asciidoc` as `devon4j-doc.pdf`
* or otherwise somehow make the `devonfw-guide` build use `devon4j-doc.asciidoc` instead and delete `master-devon4j.asciidoc`.

sjimenez77:
As far as the final file respects the documentation workflow process (level offsets, titles, etc.) for me removing one is ok. 

Just in order not to modify the [master devonfw guide](https://github.com/devonfw/devonfw-guide/blob/master/master.asciidoc) file I would suggest to maintain the `master-devon4j.asciidoc` filename independently of its contents.

hohwille:
All solved now and devon4j doc back online on website.
For the record we now have:
* devon4j-doc.asciidoc for the PDF of devon4j standalone
* master-devon4j.asciidoc for the devonfw-guide
Both files only include on other file:
* master-doc.asciidoc that includes all the sections of devon4j

Therefore we can now maintain our includes in `master-doc.asciidoc`and both `devon4j`documentation and `devonfw-guide`are in sync.
[...]

PR #193, opened 2020-01-17
Feature/160 final fix

hohwille:
Final fix for #160 as #192 broke devon4j on website/master PDF.


hohwille:
I wont wait for an official review as the website it currently broken.
I tried to fix this in devonfw-guide repo directly but figured out several drawbacks as then devon4j can not be linked as individual page anymore.

PR #194, opened 2020-01-21
WIP : added a module reporting from devonfw

vapadwal:


modules/reporting/src/main/java/com/devonfw/module/reporting/common/ReportingModuleApp.java
```diff
@@ -0,0 +1,22 @@
+package com.devonfw.module.reporting.common;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+/**
+ * This is the entry point for unit tests.
```
hohwille: If this is for Unit tests, why is it in `src/main` then rather than `src/test`?

modules/reporting/src/main/java/com/devonfw/module/reporting/common/api/Report.java
```diff
@@ -0,0 +1,102 @@
...
+public class Report<T> {
+  private String name;
+
+  private String dataSourceName;
+
+  private List<T> data;
+
+  private String templatePath;
+
+  private HashMap<String, Object> params;
```
hohwille: `HashMap` is a specific implementation. Use `Map` instead as "API".

modules/reporting/src/main/java/com/devonfw/module/reporting/common/api/Reporting.java
```diff
@@ -0,0 +1,74 @@
+package com.devonfw.module.reporting.common.api;
+
+import java.io.File;
+import java.io.OutputStream;
+import java.util.List;
+
+/**
+ * This is the interface for a simple facade to generate a report based on Jasper Reports library.
```
hohwille: IMHO this is an API that abstracts from the underlying implementation. Hence `Jasper Reports library` is "just" the current or default implementation.

modules/reporting/src/main/java/com/devonfw/module/reporting/common/api/Reporting.java
```diff
@@ -0,0 +1,74 @@
...
+   * Generates a report file with the given data, the template located in the templatePath and the type of file.
+   * specified.
+   *
+   * @param report the {@link Report} that encapsulates the data to be included in the report, the location of the
+   *        report template and the values of the parameters defined in the report template (in case template uses
+   *        parameters).
+   * @param file the file where the report must be created.
+   * @param format - report format
+   */
+  public void generateReport(Report report, File file, String format);
```
hohwille: * omit `public` keyword inside interface
* use default method delegating to generateReport with OutputStream?
* is `File` still the state of the art type of choice or should we support `java.nio.Path`?

modules/reporting/src/main/java/com/devonfw/module/reporting/common/api/Reporting.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Generates a file with a main report that stores other sub-reports.
+   *
+   * @param masterReport the main {@link Report}.
+   * @param reports a list of {@link Report} reports to be included within the main report.
+   * @param file the file where the report must be created.
+   * @param format - report format
+   */
+  public void generateSubreport(Report masterReport, List<Report> reports, File file, String format);
```
hohwille: Who. This is quite a sophisticated API. From the JavaDoc I do not really get what it will do.
I will see if I understand it when reading the implementation but JavaDoc should be more expressive.

modules/reporting/src/main/java/com/devonfw/module/reporting/common/api/Reporting.java
```diff
@@ -0,0 +1,74 @@
...
+  public void generateSubreport(Report masterReport, List<Report> reports, OutputStream stream, String format);
+
+  /**
+   * Generates a report file that contains a concatenation of reports.
+   *
+   * @param reports the list of reports to be included in the report file.
+   * @param file the file where the report must be created.
+   * @param format - report format
+   */
+  public void concatenateReports(List<Report> reports, File file, String format);
```
hohwille: What does `concat` mean in this context? If I concat two PDF files, I will get garbage.
So is the content of the reports concatenated to a single file. Is this a regular use-case. Might be cool in some cases. So I get a table of contents for the entire document similar to what we do with asciidoc and includes... Maybe the JavaDoc can again clarify the questions that obviously came to my mind here...

modules/reporting/src/main/java/com/devonfw/module/reporting/common/config/ReportExporterConfiguration.java
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Returns the {@link JRAbstractExporter} for Xls version
+   *
+   * @return
+   */
+  @Bean
+  public JRAbstractExporter getPdfExporter() {
+
+    return new JRPdfExporter() {
```
hohwille: ...
This looks like we turned a plugable concept of jasper into a monolithic approach. Does not seem to be smart to me. Am I missing something...?!? And if Jasper is providing such exporters as plugins like `JRXlsExporter` and `JRPdfExporter`, etc. why on earth do we have to override `getExporterKey()` method from outside. This appears strange to me. However, I am reviewing Jasper now while I actually should review this devon4j module instead...

hohwille: Also interesting that for `getPdfExporter` we have JavaDoc while for the others we dont. Further again this is the JavaDoc antipattern of
```
* Returns ...
* @return 
```
So simply write what the method returns after the `@return` tag withour the `Returns`. This is how JavaDoc works. I know that you already know and just took over the code from `devon` repo. However, I am seeing this a lot in reviews. Maybe we should create a better DoD or do some trainings for all the involved coders to avoid talking about basics in reviews.

PR #200, opened 2020-01-31
update link

jambulud:


documentation/architecture.asciidoc
```diff
@@ -77,7 +77,7 @@ image::images/T-Architecture.png["devonfw architecture blueprint",scaledwidth="8
 
 Please click on the architecture image to open it as SVG and click on the layers and cross-cutting topics to open the according documentation guide.
 
-We reflect this architecture in our code as described in our link:coding-conventions.asciidoc#packages[coding conventions] allowing a traceability of business components, use-cases, layers, etc. into the code and giving
+We reflect this architecture in our code as described in our link:coding-conventions[coding conventions] allowing a traceability of business components, use-cases, layers, etc. into the code and giving
```
hohwille: @jambulud thanks for your help to improve links and docs.
However, your change is wrong:
* `.asciidoc` extension must not be omitted here as otherwise you break the link on the github page while `docgen` and our wiki-travis-magic is perfectly capable to transform this and remove the extension where needed. So the link worked and here you can see that with your change it is broken and leads to 404: https://github.com/devonfw/devon4j/blob/17a7416e06b3847b826c523cd7674029def081d9/documentation/architecture.asciidoc
* removing the anchor `#packages` is also IMHO loosing valuable information. However, as an update/improvement we could change it to `#architecture-mapping`.

Can you give a rationale for this change? For me this does not make sense.

jambulud: I changed it because afte the generation of the website, I end with a link like:
`coding-conventions.html#packages`

When the correct link should be:
`coding-conventions.asciidoc.html#coding-conventions.asciidoc_packages`

If I write link:coding-conventions[coding conventions] then I get
`coding-conventions.asciidoc.html`


Check these links: 

https://devonfw.com/website/pages/docs/architecture.asciidoc.html#architecture.asciidoc_technical-architecture
image

--> clicking on the highlighted image leads to 404

The correct destination is: 
https://devonfw.com/website/pages/docs/coding-conventions.asciidoc.html#coding-conventions.asciidoc_packages

hohwille: IMHO you simply need to fix https://github.com/devonfw/devonfw-guide/issues/66
If that does not help there is another bug in https://github.com/owenh000/asciidoctor-multipage. However, we need to fix the root of the problem rather than breaking the original *.asciidoc files.

hohwille: For your understanding:
* Github code repos can render asciidoc files but require the file extension (*.asciidoc) to be present in order to make it work as the link is immediately transferred into an HTML link. Example: https://github.com/devonfw/devon4j/blob/develop/documentation/architecture.asciidoc
* Github wiki however requires the extension to be omitted. Hence the magic travis job from @sjimenez77 transforms the documentation asciidoc stuff to the wiki by removing all extensions. Example: https://github.com/devonfw/devon4j/wiki/architecture
* The `devonfw-guide` includes the wikis as submodules (to prevent overhead of cloning the code that is not required for documentation) and therefore also does not have the extensions. If that results in a 404 in the website it needs to be fixed in docgen. However, I assume that this is already done: https://github.com/devonfw/docgen/blob/master/pom.xml#L171 So either an update to the latest docgen should fix this or we have found another bug in docgen that needs to be filed, analysed, fixed, released, and tooling updated to that release.

hohwille: I do not get what is specific about this particular link. All other links follow the same conventions and should then also be broken on the website...

jambulud: docgen execution "pre-process-asciidoc" does not change between versions 4.x and 5.0.0 so I don't think that upgrading would help with the link problem

jambulud:
[devonfw-websites-wrong-links.pdf](https://github.com/devonfw/devon4j/files/4139550/devonfw-websites-wrong-links.pdf)

Here I leave a file containing wrong links found in devon4j
Maybe we can find some pattern and get the error behind this

documentation/architecture.asciidoc
hohwille: There have at least been some fixes so there is also no reason to stay on an old version.

hohwille: I meanwhile did what I suggested you to do:
* I cloned the `devonfw-guide` repo
* I ran the build
* I manually digged in the processed and transformed asciidoc files.
So here is what docgen made out of `architecture.asciidoc` (from the wiki: https://github.com/devonfw/devon4j/wiki/architecture/_compare/e6c0e92245f9b8924ca04e0b574712311e886a44):
```
...
* xref:guide-security.asciidoc[Security]
* xref:guide-logging.asciidoc[Logging]
* xref:guide-monitoring.asciidoc[Monitoring]
* xref:guide-transactions.asciidoc[Transaction-Handling]
* xref:guide-exceptions.asciidoc[Exception-Handling]
* xref:guide-i18n.asciidoc[Internationalization]
* xref:guide-dependency-injection.asciidoc[Dependency-Injection]

[[architecture.asciidoc_img-t-architecture]]
.Technical Reference Architecture
image::images/T-Architecture.png["devonfw architecture blueprint",scaledwidth="80%",align="center",link="https://devonfw.com/website/pages/docs/images/T-Architecture.svg"]

Please click on the architecture image to open it as SVG and click on the layers and cross-cutting topics to open the according documentation guide.

We reflect this architecture in our code as described in our xref:coding-conventions.asciidoc#packages[coding conventions] allowing a traceability of business components, use-cases, layers, etc. into the code and giving
developers a sound orientation within the project.
...
```

We can see that
1. xref for some strange reason even works as link to other files including other HTML files. So that is not the general problem.
2. the only difference with the "broken" link (`xref:coding-conventions.asciidoc#packages[coding conventions]`) is that it is linking to an anchor. However, the syntax of the link is perfectly correct. So this is definetly a bug in https://github.com/owenh000/asciidoctor-multipage. As we have other links with anchors I assume that another special case is required to produce the bug. IMHO this is that the actual include file is `devon4j-architecture.asciidoc` that has no section title and is therefore strangly replaced by the first included file `architecture.asciidoc` (see https://github.com/devonfw/devonfw-guide/issues/62).
So maybe a simple workaround could be to change devon4j documentation such that we move the toplevel section title from `architecture.asciidoc` to `devon4j-architecture.asciidoc` and also from `coding-conventions.asciidoc` to `devon4j-coding-conventions.asciidoc`. Should I try this, or do you have a more promising idea how to workaround this? Will you report the bug to asciidoctor-multipage? Any other ideas or action plans?

jambulud: I don't have any better idea, so you can proceed as you wish

In the meanwhile I have updated to 5.0.0

jambulud: Also, I could report a bug but it seems that the owner of the repo is not working on it anymore

https://github.com/owenh000/asciidoctor-multipage/issues/2#issuecomment-579380402

hohwille: OK. I do not know if we have ruby skilled people in devonfw but we could consider to fork and take over maintenance of the plugin if we think this is the best approach to keep for our website. We could also drop this and just use regular HTML backend of asciidoctor but I assume this will have massive drawbacks for our website...

hohwille: I have created PR #206 - I assume we have to merge and wait to see what will happen to the website or is there an easy way to test that locally for you?

hohwille: Could you formally look at the small diff of #206 and approve it for me?

jambulud: I have made a local test and I got non working links:

image

Highlighted links are not working and are referencing "code conventions"

jambulud:
I post here the results of the local tests of the website:

You can unzip the attached file and follow the steps [here](https://github.com/devonfw/devonfw.github.io#run) to run the website on your local machine (notice that it is a reduced version without devon4ng, cicdgen... in order to accelerate the build process)

documentation/architecture.asciidoc
jambulud: The asciidoc file is as follows:

image


Notice the duplicated ID. It is added by docgen https://github.com/devonfw/docgen/blob/master/pom.xml#L193

jambulud: Option? 

= General title

== Architecture

...

== Coding conventions

...

But I think that it is too much hacking

jambulud: Okay, I have tested the approach mentioned before but It is still not working: there is a link to conding conventions (coding-conventions.html#packages) but that file does not exist due to the particularities of the splitter (depending on how you include files they end in the same or in different ones)

hohwille:
My PR #206 is merged but the website is not yet updated accordingly. So we first have to check what is wrong for the website.

jambulud:
> My PR #206 is merged but the website is not yet updated accordingly. So we first have to check what is wrong for the website.

Okay but this error comes from a previous step. devonfw-guide contains devon4j changes from 6 days ago. If devonfw-guide does not get updated, neither will the website

hohwille:
Changes are now reflected on the website. However, the link in question is still broken. I am quite clueless what to do here except for finding and fixing the bug in `asciidoctor-multipage`.

maybeec:
just checked, the link we are talking here about is fixed by the latest docgen updates

Issue #147, opened 2019-09-16
Switch from dozer to orika


hohwille:
As discussed in #72 we shoud finally replace dozer with orika (or if someone comes up with a more suitable replacement and the according support it may also be something else).
However, dozer is slow and more modern solutions like orika with runtime bytecode generation can improve performance massively (bean mapping ~10x faster).

hohwille:
IMHO we should do the following:
* provide the BeanMapper API still in the current module `bean-mapping`
* offer two implementation modules `bean-mapping-dozer` and `bean-mapping-orika` each with according implementation and dependencies
* change archetype incl. config to use `bean-mapping-orika`. This will suggest new projects to use Orika what is the better alternative to recommend.
* for simplicity create a migration in `devonfw-ide` that replaced `bean-mapping` dependency to `bean-mapping-dozer`. This is trivial and lets every project with the freedom to decide to migrate or not.
* extend the bean-mapping guide with some information about this.

hohwille:
So we only need to complete and merge PRs and can then close this issue.

hohwille:
Only remaining issue is update and merge of PR for devonfw-ide.

hohwille:
PRs merged so closing as all is done.

PR #203, opened 2020-02-02
switched to orika

vapadwal:
Fix for issue https://github.com/devonfw/devon4j/issues/147

Pending items devon-ide migration and documentation


modules/beanmapping-dozer/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+  <version>${devon4j.version}</version>
+  <packaging>jar</packaging>
+  <name>${project.artifactId}</name>
+  <description>Minimal shim for bean mapping to convert between compatible Java beans (e.g. JPA entity to transfer-object and vice versa).</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.devonfw.java.modules</groupId>
+      <artifactId>devon4j-basic</artifactId>
+    </dependency>
```
hohwille: Just a nice to have, but this is a transitive dependency of `devon4j-beanmapping` so you can skip it here.

modules/beanmapping-orika/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+  <version>${devon4j.version}</version>
+  <packaging>jar</packaging>
+  <name>${project.artifactId}</name>
+  <description>Minimal shim for bean mapping to convert between compatible Java beans (e.g. JPA entity to transfer-object and vice versa).</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.devonfw.java.modules</groupId>
+      <artifactId>devon4j-basic</artifactId>
+    </dependency>
```
hohwille: Just a nice to have, but this is a transitive dependency of `devon4j-beanmapping` so you can skip it here.

modules/beanmapping-dozer/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+      <artifactId>dozer-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jaxb</groupId>
+      <artifactId>jaxb-runtime</artifactId>
+    </dependency>
```
hohwille: These dependencies came from the existing `devon4j-beanmapping` POM but do we actually know why these are needed at all for the beanmapping? Shouldnt dozer and orika come with all its dependencies to be functional? I think these dependencies are not required. Can we remove them? Or can we find out, why we need them and add some documentation as XML comment? If we need them, should they go to the api (`devon4j-beanmapping`) or to the impl (`devon4j-beanmapping-dozer` and `devon4j-beanmapping-orika`).

vapadwal: tried to completely remove these dependencies from all the modules but it threw exception while running mvn install
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.112 s <<< FAILURE! - in com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest
[ERROR] testMapEntity2Eto  Time elapsed: 2.08 s  <<< ERROR!
com.github.dozermapper.core.MappingException: 
org.xml.sax.SAXException: javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: com.sun.xml.internal.bind.v2.ContextFactory]
javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: com.sun.xml.internal.bind.v2.ContextFactory]


Have kept the above dependency in devon4j-beanmapping and removed it from other modules devon4j-beanmapping-dozer and devon4j-beanmapping-orika).

hohwille: @vapadwal thanks for testing. Do you have the stacktrace at hand so we know the reason and can add it as XML comment?
Can you resolve the other review comment so we can complete and merge this?

hohwille: I assume that this is a bug of dozer only that does not declare these XML dependencies properly. This will maybe work in older Java versions (<=1.8) as they ship with JAXB included but since Java9 it has been removed from JDK. If that is the case, we would only need the dependency for dozer and orika does not need it.

vapadwal: Thats correct only dozer has this problem, orika works

modules/beanmapping-orika/pom.xml
vapadwal: done

modules/beanmapping-dozer/pom.xml
vapadwal: done

modules/beanmapping-dozer/pom.xml
```diff
@@ -15,18 +15,49 @@
...
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.getBeanMapper(BeanMapperImplDozerTest.java:37)
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.testMapEntity2Eto(BeanMapperImplDozerTest.java:51)
+				Caused by: javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.getBeanMapper(BeanMapperImplDozerTest.java:37)
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.testMapEntity2Eto(BeanMapperImplDozerTest.java:51)
+				Caused by: java.lang.ClassNotFoundException: com.sun.xml.internal.bind.v2.ContextFactory
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.getBeanMapper(BeanMapperImplDozerTest.java:37)
+				  at com.devonfw.module.beanmapping.common.impl.dozer.BeanMapperImplDozerTest.testMapEntity2Eto(BeanMapperImplDozerTest.java:51)
+				  
+				  jaxb-api and jaxb-runtime dependencies has been added due to above error
```
hohwille: just wanted to see the trace to understand what is going on. So my guess was correct.
For XML Comment just write something like:
```
<!-- Dozer requires JAX-B what is not included in JDK since version 9+ -->
```

modules/beanmapping-dozer/pom.xml
```diff
@@ -15,18 +15,49 @@
...
+     -->
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jaxb</groupId>
+      <artifactId>jaxb-runtime</artifactId>
+    </dependency>
```
hohwille: Does `jaxb-runtime` already depend on `jaxb-api` so we could even skip the first one (`jaxb-api`).

vapadwal: yes, it depends so removed jaxb-api

modules/beanmapping-dozer/pom.xml
vapadwal: Done

hohwille:
Excellent. Code is complete and ready to be merged now.
I just went through the issue #147 - the `documentation/guide-beanmapping.asciidoc` also needs some hint about this (provide the devon4j maven dependencies for dozer and orika, link to the performance blog article and explain that we recommend orika for a new projects but still support dozer). Could you do that as well?
Have we tested our mappings (`java.time.*`, `Pageable`, etc.) with Orika? I remember some surprises with this kind of stuff in jackson...

vapadwal:
> Excellent. Code is complete and ready to be merged now.
> I just went through the issue #147 - the `documentation/guide-beanmapping.asciidoc` also needs some hint about this (provide the devon4j maven dependencies for dozer and orika, link to the performance blog article and explain that we recommend orika for new projects but still support dozer). Could you do that as well?
> Have we tested our mappings (`java.time.*`, `Pageable`, etc.) with Orika? I remember some surprises with this kind of stuff in jackson...

Documentation updated, Also I have tested orika with mtsj

modules/pom.xml
```diff
@@ -16,6 +16,8 @@
     <module>logging</module>
     <module>test</module>
     <module>beanmapping</module>
+	  <module>beanmapping-orika</module>
+	  <module>beanmapping-dozer</module>
```
hohwille: Check your formatter (indendation/tabs)

modules/jpa-dao/src/test/java/com/devonfw/example/general/common/api/config/BeansOrikaConfig.java
```diff
@@ -0,0 +1,26 @@
...
+@Configuration
+@ComponentScan(basePackages = { "com.devonfw.module.beanmapping" })
+public class BeansOrikaConfig {
+
+  @Bean
+  public MapperFacade getOrika() {
+
+    MapperFactory factory = new DefaultMapperFactory.Builder().build();
+
+    MapperFacade orika = factory.getMapperFacade();
```
hohwille: This is insufficient. We need to configure the Mapping from Entity to ETO. This was done for dozer via some magic in the XML. For orika this is already solved in the JUnit test for orika.
My suggestion would be to use the same approach as for Jackson:
* In `devon4j-beanmapping-orika` we provide a base config class that configures cross-cutting stuff such as this Entity to ETO mapping.
* Our `BeansOrikaConfig` class in app-template derives from this base config class and is the place to add project specific custom mappings and even allows to override/disable/modify the defaults from the base config class.

documentation/guide-beanmapping.asciidoc
```diff
@@ -48,7 +61,7 @@ Important is that you configure all your custom datatypes as `<copy-by-reference
 </mapping>
 ----
 
-== Bean-Mapper Usage
+== Bean-Mapper Usage(Dozer/Orika)
```
hohwille: I would remove this suffix "(Dozer/Orika)". Further our conventions say not to use special characters like braces in section titles as docgen otherwise does not work properly.

documentation/guide-beanmapping.asciidoc
```diff
@@ -21,16 +21,29 @@ public UserEto mapUser(UserEntity source) {
 Therefore we are using a `BeanMapper` for this purpose that makes our lives a lot easier.
 
 == Bean-Mapper Dependency
-To get access to the `BeanMapper` we use this dependency in our POM:
+To get access to the `BeanMapper` we have to use either of below dependency in our POM:
+
+We recommend orika for new projects but still support dozer ,please see https://www.baeldung.com/java-performance-mapping-frameworks#2-orika
```
hohwille: `We started with dozer.sourceforge.net/[dozer] in devon4j and still support it. However, we now recommend https://orika-mapper.github.io/orika-docs/[orika] (for new projects) as it is much faster (see https://www.baeldung.com/java-performance-mapping-frameworks#2-orika[Performance of Java Mapping Frameworks]).`

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/config/BeansOrikaConfig.java
```diff
@@ -7,23 +7,25 @@
...
    */
   @Bean
   public MapperFacade getOrika() {
 
     MapperFactory factory = new DefaultMapperFactory.Builder().build();
-    MapperFacade orika = factory.getMapperFacade();
+    //{@link #configureCustomMapping(MapperFactory)} can be overridden as per requirements
+    MapperFacade orika = configureCustomMapping(factory).getMapperFacade();
     return orika;
   }
```
hohwille: I would also move this code to the parent class `BaseOrikaConfig`. So this class here would be empty if there is nothing to customize. Projects can however override configureCustomMapping to extend or replace the default config. WDYT?

hohwille:
@vapadwal thanks for your final improvements. I will merge this. Great work 👍

PR #206, opened 2020-02-04
Feature/200 multipage workaround

hohwille:
Naive try as workaround to fix issues for website with `asciidoctor-multipage` as disucssed in #200


hohwille:
Did not wait for review of this trivial change to not loose time. It is just an experiment to see if it helps to fix some links on the website.

PR #207, opened 2020-02-05
Use more specific exception for not found entity

maihacke:
Currently our `GenericRepository` throws an `IllegalStateException` when invoking `find(ID id)` for a not existing id.
Spring Data converts this to an `InvalidDataAccessApiUsageException` this exception is missleading in my opinion.
Spring Data offers an `EmptyResultDataAccessException` which is more specific and allows easy handling e.g. in a RestService:

```java
  @Override
  public PersonEto getPerson(long id) {

    try {
      return this.ucManagePerson.findPerson(id);
    } catch (EmptyResultDataAccessException e) {
      throw new NotFoundException();
    }
  }
```

If the proposed change here is accepted, we could even implement this in `com.devonfw.module.rest.service.impl.RestServiceExceptionFacade`.

What do you think?


hohwille:
For the record: cause `e` should be passed to new exception:
```
    } catch (EmptyResultDataAccessException e) {
      throw new NotFoundException(e);
    }
```

PR #208, opened 2020-02-06
remove unnecesary clone

jambulud:
When you are inside a travis process, you are automatically situated under the cloned repository

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
     - REPO_SOURCE="devon4j"
     - REPO_DEST="devon4j.wiki"
     - REPO_CONSOLIDATE="devonfw-guide"
-    - GH_REPO_SOURCE="github.com/${ORG}/${REPO_SOURCE}.git"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,8 +29,9 @@ jobs:
     - stage: Documentation
       if: (branch = develop) AND (type = push)
       script:
+        # Exit devon4j folder
+        - cd ..
         # Clone repositories
-        - git clone -b develop --single-branch https://${GH_REPO_SOURCE}
```
hohwille: @jambulud Excellent improvement 👍 You save resources and speed up CI this way.
Could you also remove the `GH_REPO_SOURCE` variable if we already cleanup the travis file.
Are you doing this for all repos in `devonfw`? I never wrote this stuff and assume it has been copy&pasted to all major repos.

jambulud: @hohwille Removed unused variable

I just did this for devon4ng (and now devon4j) because I saw that devonfw-guide build process was not being triggered by the updates in the documentation

Issue #39, opened 2018-11-08
Ability to log custom fields via SLF4J


hohwille:
As discussed in #36:
It would be awesome to find a way that via SLF4J we could log custom fields such as status-code or respons-time for the performance logging  would end up as separate fields in the JSON. This would be very powerful  for kibana dashboards, etc. At the moment we are producing CSV in the  message was is not so great. I will file an issue for this but we will  not be able to address this before the release.
Current result:
```json
{"timestamp":"2018-11-06T18:36:37.638+00:00","@version":"1","message":"http://localhost:8081/services/rest/security/v1/currentuser;6678;200;;","logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter","thread_name":"http-nio-8081-exec-6","level":"INFO","level_value":20000,"appname":"basic"}
```
Expected result:
```json
{"timestamp":"2018-11-06T18:36:37.638+00:00","@version":"1","message":"http://localhost:8081/services/rest/security/v1/currentuser","response-time":"6678","status-code":"200","logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter","thread_name":"http-nio-8081-exec-6","level":"INFO","level_value":20000,"appname":"basic"}
```

hohwille:
Of course it would be possible to add the values temporary via MDC but IMHO this would be hackish and only suitable if there is no other way.

maihacke:
The PR shows how to that. It's quite easy. Is there a need for separate documentation? We could write some words in the logging-guide....

hohwille:
Fixed with PR #209 hence closing.

hohwille:
logging guide would be the prefect place. Does not make sense to create another guide for this.

maihacke:
> logging guide would be the prefect place. Does not make sense to create another guide for this.

Created another https://github.com/devonfw/devon4j/pull/213 for that

PR #209, opened 2020-02-10
Include performance info in separate fields

maihacke:
I changed our `PerformanceLogFilter` to show how https://github.com/devonfw/devon4j/issues/39 can be solved.

Currently the information is logged twice, once in the original form separated by ; and now additionally in separate fields. I kept the original format for backwards compatibility. If this is not required it is changed quite easily.

The new implementation will created logs like this:

```javascript
{"timestamp":"2020-02-10T14:36:54.299+00:00","message":"http://localhost:8081/services/rest/binary/v1/binaryobject/;21;200;;","logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter","thread_name":"http-nio-8081-exec-7","level":"INFO","url":"http://localhost:8081/services/rest/binary/v1/binaryobject/","duration":21,"statuscode":200,"errorClass":"","errorMessage":"","appname":"rootArtifactId_IS_UNDEFINED"}
```

(prettyfied)
```javascript
{ 
   "timestamp":"2020-02-10T14:36:54.299+00:00",
   "message":"http://localhost:8081/services/rest/binary/v1/binaryobject/;21;200;;",
   "logger_name":"com.devonfw.module.logging.common.impl.PerformanceLogFilter",
   "thread_name":"http-nio-8081-exec-7",
   "level":"INFO",
   "url":"http://localhost:8081/services/rest/binary/v1/binaryobject/",
   "duration":21,
   "statuscode":200,
   "errorClass":"",
   "errorMessage":"",
   "appname":"rootArtifactId_IS_UNDEFINED"
}
```

### DIFF
--- a/modules/logging/pom.xml
+++ b/modules/logging/pom.xml
@@ -19,6 +19,10 @@
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
     </dependency>
+		<dependency>
+		  <groupId>net.logstash.logback</groupId>
+		  <artifactId>logstash-logback-encoder</artifactId>
+		</dependency>    
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>jcl-over-slf4j</artifactId>
--- a/modules/logging/src/main/java/com/devonfw/module/logging/common/impl/PerformanceLogFilter.java
+++ b/modules/logging/src/main/java/com/devonfw/module/logging/common/impl/PerformanceLogFilter.java
@@ -91,8 +91,8 @@ public class PerformanceLogFilter implements Filter {
       errorClass = error.getClass().getName();
       errorMessage = error.getMessage();
     }
-    LOG.info("{};{};{};{};{}", v("url", url), v("duration", duration), v("statuscode", statusCode),
-        v("errorClass", errorClass), v("errorMessage", errorMessage));
+    LOG.info("{};{};{};{};{}", v("url", url), v("ms", duration), v("sc", statusCode), v("errCls", errorClass),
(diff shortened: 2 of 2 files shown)


hohwille:
BTW: we seem to mix different styles for our property names:
* `lowercase`
* `camlCase`
* `lower_underscore_case`

Can we align this? Also we should think about a balance of very short but still explanatory property names. One can question if JSON is a reasonable format for logs as it is massively causing redundancies since every log entry has the same structure but always includes the names of each property. This is kind of nuts. We should try to keep the property names short.

hohwille:
* I think `url` is perfect, we can not be shorter and still descriptive.
* `duration` is fine but maybe `ms` or `millis` would be even more precise as duration could also be in nanoseconds or seconds.
* For `statuscode` maybe `status` (or even `sc`)?
* For `errorClass` and `errorMessage` I only see abbreviations as other options so e.g. `errMsg` but in relation to `message` that is already given, it does not make much more sense...

maihacke:
> * I think `url` is perfect, we can not be shorter and still descriptive (except for chaning to `path` and omitting `http(s)://host:port/context-path` prefix to shorten log, but I would not really suggest this, maybe in some environment there are different URLs to access the same logical path).
> * `duration` is fine but maybe `ms` or `millis` would be even more precise as duration could also be in nanoseconds or seconds.
> * For `statuscode` maybe `status` (or even `sc`)?
> * For `errorClass` and `errorMessage` I only see abbreviations as other options so e.g. `errMsg` but in relation to `message` that is already given, it does not make much more sense...

Good hint, I changed the names accordingly

Issue #211, opened 2020-02-14
archetype: security config broken


hohwille:
Our archetype was updated from `access-control-schema.xml` to `ApplicationAccessControlConfig`. However, this change was never properly tested and we seem to deliver a broken archetype since some releases. I noticed because I deleted the obsolete `access-control-schema.xml` and that broke some tests.

hohwille:
Problem is in `WebSecurityBeansConfig` that still configures the old `AccessControlProvider` and `AccessControlSchemaProvider`.

PR #212, opened 2020-02-14
#211: fixed spring config for access-control

hohwille:
Fix for #211


hohwille:
As the build is currently broken, I will not wait for a review. If someone quickly can review, that would be great, otherwise I will merge anyhow...

Issue #90, opened 2019-05-20
 Authorization of batches


vapadwal:
Batch still needs to be started with the context of a user (e.g. our envers auditing will require a user to be logged in and that could be triggered from a batch). 

Please refer PR -https://github.com/devonfw/devon4j/pull/89

maihacke:
I implemented an example which demonstrates how to do authentication: 
https://github.com/maihacke/my-thai-star/pull/1

The solution is quite simple.
1. Provide username and password as command line arguments
2. the batch includes a tasklet steps which internally creates a usernamepasswordauthentication for theses values
3. spring security does the normal authentication. This is quite similar to what https://github.com/devonfw/devon4j/blob/develop/modules/security/src/main/java/com/devonfw/module/security/common/impl/rest/JsonUsernamePasswordAuthenticationFilter.java does.

How would we include this into devonfw.
We could try do implement tasklets which does authentication out of the box, but for that we have to make many assumptions how the authentication for the user works.

My proposal would be to just document how to do it. In the end thats less than 20 lines of codes.

What do you think @hohwille , @vapadwal ?

hohwille:
> My proposal would be to just document how to do it. In the end thats less than 20 lines of codes.

Perfect for me. Further in some cases the batch may not be able to verify any credentials (e.g. if project uses SPNEGO there is no password). If I have physical access to the machine where I start the batch, why would I need a password on the commandline? However, where does the batch get the permissions (grantedAuthorities) from? That would need to come from IAM somehow. So I am happy to read your documentation proposal as I am also not entirely clear what is the best practice in all situations. Anyhow I think a batch needs to do some kind of login to populate the authentication to spring security so that auditing (envers), authorization, etc. is working.

hohwille:
PR #215 is merged. Should we close this issue or is there anything we need to improve?

* we have an example in MTSJ
* we have a documentation describing the problem and how to address it
* we can not provide a OOTB solution applicable to all projects and demands

Hence, I am closing this issue. PRs with improvements are always welcome.

PR #215, opened 2020-02-17
Documentation for authentication of batches

maihacke:
https://github.com/devonfw/devon4j/issues/90

### DIFF
--- a/documentation/guide-batch-layer.asciidoc
+++ b/documentation/guide-batch-layer.asciidoc
@@ -87,7 +87,7 @@ To allow an easy xref:start-batch[start of the batches] from the command line it
           </excludes>
         </configuration>
       </plugin>
-      <!-- Create bootified jar for batch execution via commmand line.
+      <!-- Create bootified jar for batch execution via command line.
            Your applications spring boot app is used as main-class.
        -->
       <plugin>
@@ -128,7 +128,7 @@ java -jar <app>-batch-<version>-bootified.jar --spring.main.web-application-type
 |Parameter |Explanation
 |`--spring.main.web-application-type=none`| This disables the web app (e.g. Tomcat) 
 |`--spring.batch.job.names=<myJob>`| This specifies the name of the job to run. If you leave this out ALL jobs will be executed. Which probably does not make to much sense.
-|`<params>`| (Optional) additonal parameters which are passed to your job
+|`<params>`| (Optional) additional parameters which are passed to your job
 |===
 
 This will launch your normal spring boot app, disables the web application part and runs the designated job via Spring Boots `org.springframework.boot.autoconfigure.batch.JobLauncherCommandLineRunner`.
@@ -160,9 +160,9 @@ For that you need some logic to start at the desired offset. There different way
 
 === Using spring batch ExecutionContext for restarts
 
-By implementing the `ÌteamStream` interface in your `ItemReader` or `ItemWriter` you may store information about the batch progress in the `ExecutionContext`. You will find an example for that in the CountJob in My Thai Star.
+By implementing the `ItemStream` interface in your `ItemReader` or `ItemWriter` you may store information about the batch progress in the `ExecutionContext`. You will find an example for that in the CountJob in My Thai Star.
 
-Addtional hint: It is important that bean definition method of your `ItemReader`/`ItemWriter` return types implementing `ItemStream`(and not just `ItemReader` or `ItemWriter` alone). For that the `ItemStreamReder` and `ItemStreamWriter` interfaces are provided.
+Additional hint: It is important that bean definition method of your `ItemReader`/`ItemWriter` return types implementing `ItemStream`(and not just `ItemReader` or `ItemWriter` alone). For that the `ItemStreamReader` and `ItemStreamWriter` interfaces are provided.
 
 == Exit codes
 
@@ -192,7 +192,7 @@ public class SpringBootApp {
 Spring batch uses several database tables to store the status of batch executions.
 Each execution may have link:https://docs.spring.io/spring-batch/docs/current/reference/html/index-single.html#jobexecution[different status].
 You may use this mechanism to link:https://docs.spring.io/spring-batch/docs/current/reference/html/index-single.html#stoppingAJob[gracefully stop batches].
-Additonally in some edge cases (batch process crashed) the execution status may be in an undesired state.
+Additionally in some edge cases (batch process crashed) the execution status may be in an undesired state.
 E.g. the state will be running, despite the process crashed sometime ago.
 For that cases you have to change the status of the execution in the database.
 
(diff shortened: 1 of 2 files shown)


documentation/guide-batch-layer.asciidoc
```diff
@@ -212,8 +212,70 @@ Show help::
 As you can the each invocation includes the jdbc connection string to your database.
 This means that you have to make sure that the corresponding DB driver is in the classpath (the prepared jar only contains H2).
 
+== Authentication
+
+Most business application incorporate authentication and autorisation.
```
hohwille: `autorisation` should be `authorization` ;)

documentation/guide-batch-layer.asciidoc
```diff
@@ -212,8 +212,70 @@ Show help::
...
+Your spring boot application will implement some kind of security, e.g. integrated login with username+password or in many cases authentication via an existing IAM.
+For security reasons your batch should also implement an authentication mechanism and obey the authorization implemented in your application (e.g. via @RolesAllowed).
+
+Since there are many different authentication mechanism we cannot provide an out-of-the-box solution in devonfw, but we describe a pattern who this can be implemeneted in devonfw batches.
```
hohwille: `who` should be `how`

documentation/guide-batch-layer.asciidoc
```diff
@@ -212,8 +212,70 @@ Show help::
...
+public class SimpleAuthenticationTasklet implements Tasklet {
+
+  @Override
+  public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
+
+    String username = chunkContext.getStepContext().getStepExecution().getJobParameters().getString("username");
+    String password = chunkContext.getStepContext().getStepExecution().getJobParameters().getString("password");
+    Authentication authentication = new UsernamePasswordAuthenticationToken(username, password);
+
+    SecurityContextHolder.getContext().setAuthentication(authentication);
```
hohwille: wont this overrule spring-security so the password is never checked? Or is there some magic in spring-batch, I am not aware of?

documentation/guide-batch-layer.asciidoc
```diff
@@ -212,8 +212,70 @@ Show help::
...
+  }
+...
+----
+
 == Tipps & tricks
 
 === Identifying job parameters
 
-String uses a jobs parameters to identify https://docs.spring.io/spring-batch/docs/current/reference/html/domain.html#jobexecution[job executions]. Parameters starting with "-" are not considered for identifying a job execution.
+Spring uses a jobs parameters to identify https://docs.spring.io/spring-batch/docs/current/reference/html/domain.html#jobexecution[job executions]. Parameters starting with "-" are not considered for identifying a job execution.
```
hohwille: Nice catch. I did the same mistake already several times.

documentation/guide-logging.asciidoc
```diff
@@ -111,4 +111,34 @@ In order to correlate separate HTTP requests to services belonging to the same u
 == Monitoring
 In highly distributed systems (from clustering up to microservices) it might get tedious to search for problems and details in log files. Therefore, we recommend to setup a central log and analysis server for your application landscape. Then you feed the logs from all your applications (using http://logstash.net/[logstash]) into that central server that adds them to a search index to allow fast searches (using http://www.elasticsearch.org/[elasticsearch]). This should be completed with a UI that allows dashboards and reports based on data aggregated from all the logs.
 This is addressed by https://www.elastic.co/webinars/introduction-elk-stack[ELK] or https://www.graylog.org/[Graylog].
- 
\ No newline at end of file
+
+=== Adding custom values to JSON log
+
+When you use a external system for gathering distrubited logs, we strongly suggest that you use a JSON based log format (e.g. by using the provided `logback.xml`, see above).
```
hohwille: Do you know why you mix features in your PRs several times? Something is odd with github? Or the way you manage your feature-brances? Or the way you merge?

maihacke: It is not a general problem there were conflicts because I started with this task here before the PR for the batch topic itself was merged. Sadly I put the minor change for logback on by develop branch (instead of creating a feature branch). That mixed everything up. I am ashame...

I removed that commit from the PR.

documentation/guide-batch-layer.asciidoc
maihacke: This is Spring Security magic. Because Authentication.isAuthenticated is `false` Spring automatically does an authentication with the provided credentials.

documentation/guide-batch-layer.asciidoc
maihacke: changed

documentation/guide-batch-layer.asciidoc
maihacke: yes!

maihacke:
Thanks for the review. I did the requested fixes and additionally a spell check (now I have a VScode plugin for it ;-))

documentation/guide-batch-layer.asciidoc
```diff
@@ -212,8 +212,70 @@ Show help::
...
+
+    SecurityContextHolder.getContext().setAuthentication(authentication);
+    return RepeatStatus.FINISHED;
+  }
+
+}
+----
+
+The username and password have to be supplied via two cli parameters `-username` and `-password`. This implementation creates an "authenticated" `Authentication` and sets in the Spring Security context. This is just for demonstration normally you should not provide passwords via commandline. The acutal authentication will be done automatically via Spring Security as in your "normal" application. 
+If you have a more complex authentication mechanism in your application e.g. via OpenID connect just call this in the tasklet. Naturally you may read authentication parameters (e.g. secrets) from the commandline or more securely from a configuration file.
```
hohwille: So what is the "best practice" how to actually do it properly? Also an application may use arbitrary mechanism to authenticate an end-user. There may not even be something like a password e.g. in contexts such as `JWT` or `Fido2`. So to be honest: This is all a valuable step to fill the gap but this is more a proposal for discussion than the final solution...

hohwille: To open the discussion, here are some thoughts:
* if credentials have to be passed whenever a batch is started, we will prevent fully automated batch job processing what is IMHO a general requirement
* if we provide the password as part of the deployment (via config file or so), we IMHO weaken security. In such case it is even better to do some kind of "dummy authentication" for batches that do not need credentials.

maihacke: I don‘t think that we should or can provide a general solution here. The solution should be part of a security concept which goes far beyond what we could describe here and has to be adopted to the concrete requirements.
What we can offer is partly solutions for different authentication themes and maybe platforms in the future.
E.g. how you could securely provide secrets to your batch depends on your environment - in AWS you may prefer AWS Secrets or there maybe special hardware tokens in place, or ....

PR #219, opened 2020-02-20
#200: devonfw/docgen#18: reworked structure of devon4j doc to fix links

hohwille:
See #200 and devonfw/docgen#18


documentation/devon4j.asciidoc
```diff
@@ -7,12 +22,167 @@ The following sections contain the complete compendium of https://github.com/dev
 You can also read the latest version of this documentation online in the http://github.com/devonfw/devon4j/wiki[devon4j wiki]
 or at https://devonfw.com/website/pages/docs/master-devon4j.asciidoc.html[devon4j on devonfw.com].
 
-include::devon4j-architecture.asciidoc[leveloffset=1]
+== Architecture
 
-include::devon4j-coding.asciidoc[leveloffset=1]
+include::architecture.asciidoc[leveloffset=1]
```
jambulud: If you want to have the contents of "architecture.asciidoc" under "== Architecture" then you have to  include it with leveloffset=2

A rule  of thumb is, if you want an include file to be under a section, include it using the number of "=" as "leveloffset"

Example:

= level 0

include::file.asciidoc[leveloffset=1]  // 1 "=" then leveloffset=1

== level 1

include::file.asciidoc[leveloffset=2]  // 2 "=" then leveloffset=2


If you do:

== level 1

include::file.asciidoc[leveloffset=1] 

You will end with an empty section

level 1
// here nothing
included file (also on level one due to leveloffset)
// here text

documentation/devon4j.asciidoc
```diff
@@ -7,12 +22,167 @@ The following sections contain the complete compendium of https://github.com/dev
...
-include::devon4j-layers.asciidoc[leveloffset=1]
+<<<<
 
-include::devon4j-guides.asciidoc[leveloffset=1]
+include::guide-component.asciidoc[leveloffset=2]
 
-include::devon4j-tutorials.asciidoc[leveloffset=1]
+== Coding Conventions
+
+include::coding-conventions.asciidoc[leveloffset=1]
```
jambulud: Here the same situation as in architecture section

sjimenez77:
Ples @hohwille, check @jambulud comments. Thanks!

documentation/devon4j.asciidoc
hohwille: Thanks for your review but I think this is a missunderstanding. As a workaround for the strange behaviour of the devonfw website with `asciidoctor-mutlipage` we had to remove the toplevel section from `architecture` and `coding-conventions`.
Look at the contents of the included files: There is no `=` section but all sections start with `==` (also for asciidoc it is discouraged to have multiple level 0 `=` sections in the same file).
So IMHO your comment your concern is invalid.
BTW: If I change as suggested and test the documentation build asciidoc is complaining about invalid section numbering while currently it does not.

hohwille:
@sjimenez77 I did, can I merge, or do we need to block this any longer?

Issue #44, opened 2018-11-30
Consider imporvements for monitoring


hohwille:
For monitoring of a devonfw documentation there should be more guidance and features.
* https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready[spring-boot actuator]
* https://github.com/codecentric/spring-boot-admin[spring-boot-admin]
* http://messadmin.sourceforge.net/[messadmin]

hohwille:
See also here:
https://github.com/devonfw-wiki/devon4j/wiki/guide-apm
(considering JavaMeldoy)

sjimenez77:
Sometime ago we already created a demo and a cookbook entry in the devonfw guide for the integration of Spring Boot Admin https://github.com/devonfw/devon/wiki/Spring-boot-admin-Integration-with-devon4j. The document is probably deprecated, but could be a starting point.

nricheton:
Hello, 

Following a discussion w/ Jörg & Santos, here is my input on monitoring. 

Overview : 

- Monitoring should be free / effortless, should be set up on day 1.
- A core set of monitoring should be available from dev environment to production 
- You should not need to ask a developer/an expert to retrieve or interpret the metrics/monitoring data. This info should be clear, have a first layer of analysis and tell that situation is OK or directly point out the issues.

Some examples of questions that should have an **immediate** answer from a monitoring solution. 
(Immediate = display a web page, now)
- Is the right version deployed ?
- Was it able to start successfully ?
- From the app/module/service point of view, is connectivity OK ?
- From the app/module/service point of view, is configuration OK ?
- From the app/module/service point of view, is performance OK ?
- From the app/module/service point of view, is memory OK ?
- From the app/module/service point of view, is data OK ?
- From the app/module/service point of view, is number of runtime business or technical error acceptable ?
- What is the average response time of X (service, operation, data request, business rule) ?
- If not, what should I do ?
- What is the percentage/number of technical errors of X (service, operation, data request, business rule) ?
- What is the percentage/number of business errors of X (service, operation, data request, business rule) ?
- Is caching efficient in X (service, operation, data request, business rule) ?
- Is there background/async process running, when will it finish ?
- Is there background/async process scheduled and does scheduling work ?
- Is there any issue in background processing, which business data is causing issue ?

All these answers are priceless in production, but even in development/testing environments, where they are a clear indicator of the upcoming issues in the next stage. 

[...]

PR #221, opened 2020-02-24
Wrote monitoring guide

maihacke:
see #44


hohwille:
thanks for the doc improvement 👍 I did the review constructive while merging...
Still room to improve but as we had only TODOs in the guide before I merged without wasting any more time...

PR #229, opened 2020-04-02
Introduce type params for Kafka API

maihacke:


maihacke:
Could any one of you please do a short review. It is just a minor change to polish the API a bit.

PR #230, opened 2020-04-02
Provide starters for kafka

maihacke:
@ravicm2 I created two starters for kafka.
Please use them in the example app you are currently working on.

### DIFF
--- a/boms/minimal/pom.xml
+++ b/boms/minimal/pom.xml
@@ -186,6 +186,16 @@
         <artifactId>devon4j-starter-spring-data-jpa</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.devonfw.java.starters</groupId>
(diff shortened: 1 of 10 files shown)


maihacke:
Build has been fixed

hohwille:
BTW - a general feedback: We have already merged PRs with the kafka feature but so far I can not find any documentation guide for it. Usually we try to keep documentation and code in sync with PRs. Would be nice if we can include some kafka documentation untill we do the 2020.04 release as we should not release a feature without documentation.

maihacke:
> @maihacke great that you also added startes to make usage more easy. 👍
> To me the usage of spring.factories, etc. looks reasonable and I hereby approve. However, we would need an example app in `devonfw-sample` using these starters in action so we can also verify that this is really working as expected (hard to tell just from reading the code in the browser).

I included a test in the starter-sender, which checks, that the configuration is working technically. The starter-receiver, will be used in the example.

PR #232, opened 2020-04-03
Rework generics in devon-kafka

maihacke:
Made some improvements to the generics in devon-kafka


maihacke:
> @maihacke thanks for the improvement 👍
> Makes sense that the API holds the generics so users can use it type-safe for their custom keys and messages but the reusable implementations simply bind the generics to Object as they are not requesting any specific type (e.g. for error handling).
> I can not say why this `tracer` was removed but seems all reasonable to me so ready for merge.

Thanks for the review. I removed the tracer because it wasn't used.

PR #238, opened 2020-04-16
Jwtbr needs to merged in develop

vapadwal:


hohwille:
I will have a closer look and do some final improvements myself. I am not yet sure when I will find a day of time to get this done.

hohwille:
We would really need an example demostrating this feature in action. I added some fundamental tests but do not have the time to complete the tests right now (already way to late).
* How should `JwtLoginFilter` be used? Is this actually best practice? To send login+password via proprietary JSON IMHO seems to be against regular HTTP Authentication standards.
* We need a full fledget sample or JUnit subsystem test with RANDOM_PORT testing the authentication via JWT as OAuth Bearer token with `JwtAuthenticationFilter`.
* Dont we need a spring-starter for this feature?
* Where is the documentation for all this stuff in the documentation folder?

hohwille:
@vapadwal how do we proceed? Should I merge this and you create another PR for improvements or will you create a PR for this PR before we merge?

vapadwal:
> @vapadwal how do we proceed? Should I merge this and you create another PR for improvements or will you create a PR for this PR before we merge?

both plans sounds good, but it would be better if you merge this PR first

PR #239, opened 2020-04-16
Composite Keys changes

isandesh1986:


maybeec:
I think, we should not provide these classes in archetype, but in the persistence module where the other abstract super classes also are located.

hohwille:
I agree with @maybeec but what we could improve in devon4j is that AbstractEto could be maked somewhat more reusable for this case. E.g. we could expose another abstract super-class using a generic for ID and then update our bean-mapping to support this on the new abstract super-class so our magic mapping with `modificationCounter` will also work in case a composite key is used. WDYT?

PR #245, opened 2020-04-23
update jwt

vapadwal:
identified few issues while integrating with mtsj , please see mtsj PR for the same https://github.com/devonfw/my-thai-star/pull/334


modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/impl/JwtManagerImpl.java
```diff
@@ -45,7 +48,7 @@ public JwtManagerImpl() {
   public Jws<Claims> decode(String jwt) {
 
     PublicKey key = this.keyStoreAccess.getPublicKey(this.jwtConfig.getAlias());
-    Jws<Claims> token = Jwts.parser().setSigningKey(key).parseClaimsJws(jwt);
+    Jws<Claims> token = Jwts.parser().setSigningKey(key).parseClaimsJws(jwt.replace(JwtConstants.TOKEN_PREFIX, ""));
```
hohwille, line 51: I am not sure about this. Actually the provided `jwt` argument should be the JWT itself. The code I provided already removes the "Bearer" prefix elsewhere. Surely it wont hurt to also do it here but what if there are multiple whitespaces around or the Upper/Lower-Case does not match the TOKEN_PREFIX. Therefore it would IMHO be better to have only one place in the code that does this. If you want to reuse here, you could expose as static mehtod in the other impl and reuse here... WDYT?

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/impl/JwtManagerImpl.java
```diff
@@ -25,6 +27,7 @@
  *
  * @since 2020.04.001
  */
+@Named
```
hohwille, line 30: My vision was that we also create a spring-boot starter.
The `@Named` only helps if the user does a component-scan in this package within its spring-config.
A spring-starter would therefore much better.
Anyhow good that you readded this annotation here. It can only help but not hurt! So thanks for this fix.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/impl/JwtManagerImpl.java
vapadwal: I think you are right, problem is that i test this with mtsj. while debugging i found , there multipe (two) Bearer in token 

Bearer Bearer eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJ3YWl0ZXIiLCJyb2xlcyI6IlJPTEVfV2FpdGVyLG10cy5GaW5kQm9va2luZyxtdHMuRmluZE9yZGVyIiwiaXNzIjoiZGV2b25mdyIsImlhdCI6MTU4NzY1OTg5NywibmJmI-----------

basically mtsj angular adds this extra Bearer keyword to the token

latMap((token) => {
        const authReq = !!token
          ? req.clone({
              setHeaders: { Authorization: 'Bearer ' + token },
            })

I will remove the code (jwt.replace(JwtConstants.TOKEN_PREFIX, "")) from JwtManagerImpl

hohwille: Than angular client also is correct. The token itself should not contain `Bearer`. Maybe the code to extract the token on the client is broken or I did something wrong in JwtLoginFilter so we added `Bearer` prefix twice (or maybe it does not belong to response header at all)? This part is new to me, however I did not intend to change anything of this behaviour from your code. Maybe it was wrong before or I messed something during refactoring.

hohwille:
> In documentation
> 
> ```
> <dependency>
> 	<groupId>com.devonfw.java.modules</groupId>
> 	<artifactId>devon4j-starter-security-jwt</artifactId>
> </dependency>
> ```
> 
> it should be
> 
> ```
> <dependency>
> 	     <groupId>com.devonfw.java.starters</groupId>
>         <artifactId>devon4j-starter-security-jwt</artifactId>
>     </dependency>
> ```
> 
> Also
> `http.addFilterBefore(getJwtAuthenticationFilter(), UsernamePasswordAuthenticationFilter.class);`
> 
> it should be getJwtLoginFilter()

You can fix this directly next time.
I just comitted the changes to your fork.
Thanks for the review feedback and sorry for my mistakes. Too much hussle at the moment. So great that with our review process we find such issues.

Issue #241, opened 2020-04-21
Add module to support JWT and parts of OAuth


hohwille:
As a devon4j user, I want to use authentication via JWT so that my microservices do not implement the actual authentication mechanism itself but can instead only validate a JWT from an access-manager like keycloak (or as alternative provided by the devon4j app itself). This will also by itself already prevent CSRF and improve security. Further it can free the servers/apps/microservice from any state and HTTP session to make them more scaleable.

hohwille:
For the record we had multiple PRs and I created this issue to make this more transparent for our changelog. PRs are:
* #51
* #76
* #154
* #157
* #223 
* #238

hohwille:
Things to consider or to complete for the release:
* [ ] create documentation (`guide-jwt` linked from `guide-access-control`)
* [ ] improve tests
* [ ] create spring boot starter for JWT
* after the release demonstrate the feature by adopting MTSJ and/or creating a `devonfw-sample`. If possible test with SNAPSHOT before release...

hohwille:
Done and merged.

PR #247, opened 2020-04-24
#241: JWT: added spring-boot starters and auto-config, added JUnit tests

hohwille:
Fix for #241:
* added auto-configs
* added spring-boot starters
* added JUnit tests


hohwille:
> [ERROR] /home/travis/build/devonfw/devon4j/starters/starter-security-jwt/src/test/java/com/devonfw/module/security/jwt/common/api/JwtCreatorTest.java:[31,11] error: cannot find symbol
  symbol:   method of(String,String)
  location: interface Set

Method I used is not available in JDK 1.8.
Will fix asap.

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/base/JwtAuthenticationFilter.java
```diff
@@ -21,6 +24,8 @@
  *
  * @since 2020.04.001
  */
+@Component
+@Order(SecurityProperties.BASIC_AUTH_ORDER - 1)
```
maihacke, line 28: Nice ;-)

hohwille: Trade-off: to make spring-boot starter pluggable with convention over configuration I had to find a solution. Projects are free to depend on `devon4j-security-jwt` instead of the starter and manually configure the beans (e.g. creating the filter with `new` in WebSecurityConfigurer and add to filter chain in the desired place).

hohwille:
> Please change
> 
> https://github.com/devonfw/devon4j/blob/67e7aed7629caf52eb0c2d0de23d1d62bb8167b7/modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/impl/JwtConfigProperties.java#L187
> . It should be delay not "deplay".
> 
> Please also fix this in https://github.com/vapadwal/devon4j/blob/update_jwt/documentation/guide-jwt.asciidoc#configuration

OMG stupid typo. That's what reviews are good for :) Thanks so much @maihacke

modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/base/JwtAuthenticationFilter.java
vapadwal: Can we define @Component here as in JwtAutoConfiguration we have defined it as Bean
 @Bean
  public JwtAuthenticationFilter jwtAuthenticationFilter() {

    return new JwtAuthenticationFilter();
  }

Same for 
 @Bean
  public JwtConfigProperties jwtConfigProperties() {

    return new JwtConfigProperties();
  }

Getting error while running mtsj
The bean 'jwtConfigProperties', defined in class path resource [com/devonfw/module/security/jwt/common/impl/JwtAutoConfiguration.class], could not be registered. A bean with that name has already been defined in URL [jar:file:/D:/projects/devon/devon-ide-scripts-3.0.0-beta24/conf/.m2/repository/com/devonfw/java/modules/devon4j-security-jwt/2020.04.001-SNAPSHOT/devon4j-security-jwt-2020.04.001-SNAPSHOT.jar!/com/devonfw/module/security/jwt/common/impl/JwtConfigProperties.class] and overriding is disabled.

same error for JwtAuthenticationFilter

PR #248, opened 2020-04-27
#241: jwt documentation update

vapadwal:
documentation enhancement for #241


vapadwal:
> This filter is automatically added via auto-configuration. IMHO this change/PR is therefore wrong.
> Doesn't this work without this change in MTSJ?

Yes I was also wondering that, might be some thing to do with mtsj. but it works when i add that jwtAuthenticationFilter in addFilterBefore

basically it gives some security error.
2020-04-27 15:43:07.429 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.w.u.matcher.AntPathRequestMatcher  : Checking match of request : '/services/rest/security/v1/currentuser/'; against '/services/rest/ordermanagement/v1/order/cancelorder/**'
2020-04-27 15:43:07.429 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.w.u.matcher.AntPathRequestMatcher  : Request 'GET /services/rest/security/v1/currentuser/' doesn't match 'POST /login'
2020-04-27 15:43:07.429 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.w.a.i.FilterSecurityInterceptor    : Secure object: FilterInvocation: URL: /services/rest/security/v1/currentuser/; Attributes: [authenticated]
2020-04-27 15:43:07.429 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.w.a.i.FilterSecurityInterceptor    : Previously Authenticated: org.springframework.security.authentication.AnonymousAuthenticationToken@98f946e9: Principal: anonymousUser; Credentials: [PROTECTED]; Authenticated: true; Details: org.springframework.security.web.authentication.WebAuthenticationDetails@b364: RemoteIpAddress: 0:0:0:0:0:0:0:1; SessionId: null; Granted Authorities: ROLE_ANONYMOUS
2020-04-27 15:43:07.440 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.access.vote.AffirmativeBased       : Voter: org.springframework.security.web.access.expression.WebExpressionVoter@508bce31, returned: -1
2020-04-27 15:43:07.444 DEBUG 18364 --- [nio-8081-exec-4] o.s.b.a.audit.listener.AuditListener     : AuditEvent [timestamp=2020-04-27T10:13:07.441841300Z, principal=anonymousUser, type=AUTHORIZATION_FAILURE, data={details=org.springframework.security.web.authentication.WebAuthenticationDetails@b364: RemoteIpAddress: 0:0:0:0:0:0:0:1; SessionId: null, type=org.springframework.security.access.AccessDeniedException, message=Access is denied}]
2020-04-27 15:43:07.446 DEBUG 18364 --- [nio-8081-exec-4] o.s.s.w.a.ExceptionTranslationFilter     : Access is denied (user is anonymous); redirecting to authentication entry point

org.springframework.security.access.AccessDeniedException: Access is denied

hohwille:
From the spring documentation the two annotations I added should actually add the filter automatically to the right position:
https://github.com/devonfw/devon4j/blob/develop/modules/security-jwt/src/main/java/com/devonfw/module/security/jwt/common/base/JwtAuthenticationFilter.java#L28

So its great that you give the new JWT feature a real test. As it seems your realistic test is revealing that this is not working as expected. So I either need to dig more or we have remove that configuration (replace the 2 current annotations with just `@Named`) and go with your approach. My intention is always to make usage as easy and plugable as possible with spring-boot starters. But if it does not work, we should change the documentation and explain how to do it manually.

However, I separated two aspects in the documentation as `JwtLoginFilter` is a very special case that might only be used by my-thai-star but is not recommended for every app. So still this PR is not right, yet.

documentation/guide-jwt.asciidoc
```diff
@@ -64,19 +64,28 @@ To allow a client to login with username and password to get a JWT for sub-seque
 
 [source,java]
 ----
+  
   @Bean
-  public JwtLoginFilter getJwtLoginFilter() throws RuntimeException {
+  public JwtLoginFilter getJwtLoginFilter() throws Exception {
```
hohwille, line 69: I still do not see why we need this throws declaration at all here, but not so important.

documentation/guide-jwt.asciidoc
```diff
@@ -64,19 +64,28 @@ To allow a client to login with username and password to get a JWT for sub-seque
...
 
     JwtLoginFilter jwtLoginFilter = new JwtLoginFilter("/login");
     jwtLoginFilter.setAuthenticationManager(authenticationManager());
     jwtLoginFilter.setUserDetailsService(this.userDetailsService);
     return jwtLoginFilter;
   }
   
+  @Bean
+  public JwtAuthenticationFilter getJwtAuthenticationFilter() throws Exception {
```
hohwille, line 78: Also throws Exception seems pointless here.

hohwille:
```
@Bean
public JwtAuthenticationFilter getJwtAuthenticationFilter() {
  return new JwtAuthenticationFilter();
}
```

documentation/guide-jwt.asciidoc
vapadwal: here authenticationManager() throws exception

hohwille: OK. Then here it is required. The getJwtAuthenticationFilter() does not throw an exception, however...

Issue #258, opened 2020-05-04
archetype is still lacking a .gitignore


hohwille:
As a devon4j user, I want to get a `.gitignore` with reasonable defaults for a Java project whenever I create one from the archetype so that I can directly commit and push to some git repository.

hohwille:
We should simply copy our existing https://github.com/devonfw/devon4j/blob/develop/.gitignore to the archetype and ensure that it is included.
IMHO the copy can be created here:
https://github.com/devonfw/devon4j/tree/develop/templates/server/src/main/resources/archetype-resources
This should be very easy to do. Maybe a config has to be set in the pom so the file is actually included (typically `.*` files are IMHO excluded).

Acceptance test: when a new java project is created from the archetype it includes such a `.gitignore` file.

hohwille:
Done with PR #263

PR #263, opened 2020-05-13
Added .gitignore in archtype

vapadwal:
Please see https://github.com/devonfw/devon4j/issues/258


templates/server/pom.xml
```diff
@@ -37,6 +37,8 @@
           <delimiters>
             <delimiter>$[*]</delimiter>
           </delimiters>
+          <!--Requires below statement, please see https://issues.apache.org/jira/browse/MRESOURCES-190 -->
+          <addDefaultExcludes>false</addDefaultExcludes>
```
hohwille, line 41: We need to test this. The explicit `include` is not enough? Disabling the default excludes may include some other undesired files into the archetype. Maybe we can create archetype before and after this change and diff the files included.

vapadwal: yes have ask @Rohit90700  also to test this at his end

Rohit90700: Hi @vapadwal  , I have tested ,its working fine for me.

hohwille: @Rohit90700 thanks for testing so far.
I am not in doubt that with the change `.gitignore` will be added but will also other undesired files (`*.bak`, `*.log`,`.project`, ...) be added accidentally in the archetype?
Did you diff the outcome before and after this change as suggested in my comment above?

Rohit90700: Hi @hohwille , I have tested both scenario before and after the change as you suggested above comment. there is only one file difference of .gitignore .

hohwille:
Ready to be merged as soon as updated and green.

Issue #254, opened 2020-04-28
Kafka: Integrate authentication


maihacke:
The current kafka module for devon4j does not provide an authentication mechanism out of the box.
A good starting point could be to integrate JWT authentication.
For producers the implementation could automatically our at least assist with putting the token into a kafka header.
For consumers a "plugin" for spring security could be provided with extract the token from the header and initializes spring security accordingly.

maihacke:
https://github.com/devonfw/devon4j/pull/264

PR #264, opened 2020-05-19
JWT Authentication support for devon4j-kafka

ravicm2:
Add JWT authentication (see https://github.com/devonfw/devon4j/pull/264)


maihacke:
In a discussion with @hohwille we decided to move the jwt-authentication for kafka into devon4j-security-jwt. So if people are using kafka without jwt the do not have dependencies for jwt at all. I marked the dependencies to kafka and servlet in devon4j-security-jwt as optional and add a @contionalonclass to the autonconfiguration, so if people are using a pure kafka environment the jwt-servlet-filter should not be there. @vapadwal

maihacke:
closes https://github.com/devonfw/devon4j/issues/254

Issue #261, opened 2020-05-11
Support for JUnit4


hohwille:
As a user of devon4j, I want to upgrade my project to the latest version of devon4j to get security-fixes and new features, but I do not want to be forced to migrate to JUnit5 so that I do not have to waste lots of effort and money without adding real value for a "legacy project".

The automigration in devonfw-ide is not working properly so manual migration to JUnit5 is required. It absolutely makes sense to start greenfield projects with JUnit5 but projects that started with JUnit4 and have tons of tests already should not be forced to upgrade.
Of course this can be fiddled manually with exclusions and added dependencies. However, it would be much better to add a `devon4j-test-junit4` module that cleanly provides this support and change the migration in `devonfw-ide` to instead replace `devon4j-test` with `devon4j-test-junit4` for projects prior to devon4j `3.2.0`.

hohwille:
Done with PR #268

PR #268, opened 2020-05-26
support for junit4

vapadwal:
Support for junit4 https://github.com/devonfw/devon4j/issues/261


modules/test-junit4/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <parent>
+    <groupId>com.devonfw.java.dev</groupId>
+    <artifactId>devon4j-modules</artifactId>
+    <version>dev-SNAPSHOT</version>
+  </parent>
+  <groupId>com.devonfw.java.modules</groupId>
+  <artifactId>devon4j-test-junit4</artifactId>
+  <version>${devon4j.version}</version>
+  <name>${project.artifactId}</name>
+  <description>Module with code and configuration for tests.</description>
```
hohwille, line 13: could you update the description to include this is for junit4 backward compatibility.

modules/test/src/main/resources/logback.xml
```diff
@@ -14,7 +14,6 @@
   </root>
 
   <!-- Minimize infrastructure debug logs -->
-  <logger name="org.dozer" level="INFO"/>
```
hohwille, line 17: We migrated to a new dozer version that has a new namespace (with github in it).
As dozer unfortunately logs like hell by default (what is entirely nuts) we have to ensure that dozer is at least on info level and not on debug (otherwise the logs explode).
We can address this in a separate PR but I missed this during the dozer migration and this needs to be fixed urgently as well (mainly in the `logback.xml` of the archetype in `server` module).

hohwille: I have seen this in real-life projects where builds/releases failed due to OutOfMemoryError cause dozer was logging gigabytes of properties that have been mapped.

PR #274, opened 2020-06-03
Retry pattern for a specific topic support

ravicm2:


ravicm2:
> regarding the new props:
> 
> ```
>   messaging.retry.default.retry-policy.retryPeriod.employeeapp-employee-v1-add=3000
>   messaging.retry.default.retry-policy.retryableExceptions.employeeapp-employee-v1-add=<Class names of exceptions for which a retry should be performed>
>   messaging.retry.default.retry-policy.retryableExceptionsTraverseCauses.employeeapp-employee-v1-add=true
> ```
> 
> ...
> I think it is a bit mixed up currently. It should be possible to set values for each property or if users do not want this to set them for all topics together (as it is currently).
> 
> So the properties should be:
> 
> ```
> messaging.retry.retry-policy.retryPeriod.employeeapp-employee-v1-add
> ```
> 
> or
> 
> ```
> messaging.retry.default.retry-policy.retryPeriod.default
> ```
> 
> but not 'default' and '' at the same time in one property. If you do not need the word "default" in the property name it would be even better, because it is compatible with the current set of properties

Hi @maihacke , The current implementation(on this branch) is like if they didn't provide the topic name at the end of property , spring throws a conversion exception saying that it cant convert the string to map. 

**`Description:

Failed to bind properties under 'messaging.retry.default.retry-policy.retryable-exceptions' to java.util.Map<java.lang.String, java.util.Set<java.lang.String>>:

    Reason: No converter found capable of converting from type [java.lang.String] to type [java.util.Map<java.lang.String, java.util.Set<java.lang.String>>]

Action:

Update your application's configuration`**

With you comment i'm trying to understand that user should be able to provide either topic name at the end  of property or the key **default** which represents same value for all the topics. Am i right?

maihacke:
It should be 
messaging.retry.retry-policy.retryPeriod.default

(without default twice in the property name)

maihacke:
I approved the PR. If you update it to develop, I'll do the merge.

Issue #260, opened 2020-05-08
Support for opentracing


hohwille:
As a devon4j user, I want to have opentracing support so that I can monitor my IT landscape better and rely on open standards supported by many tools rather than proprietary correlation-ID.

In devonfw we have created the concept of the correlation-ID which is nice. However devonfw is comitted to open-standards and meanwhile something simlar and even better has be come a kind of standard called opentracing:
* https://opentracing.io/
* https://www.datadoghq.com/blog/instrument-opencensus-opentracing-and-openmetrics/

We can have a look at existing libs like jaeger but the functionally is so simple that we can implement this oursevles in the existing code.
Finally, we should provide some downward compatibility for correlation-ID, so e.g. the X-Correlation-ID header should still be honored if present and no opentracing headers are present.

hohwille:
Links to current implementation and behaviour:
* https://github.com/devonfw/devon4j/blob/c8f25cb073909db78d0329eda7ce44d93c9e9bae/modules/logging/src/main/java/com/devonfw/module/logging/common/impl/DiagnosticContextFilter.java#L90
* https://github.com/devonfw/devon4j/blob/9b8f9461997bb5d8048a9fec42bec3625652eec3/modules/logging/src/main/java/com/devonfw/module/logging/common/impl/DiagnosticContextFacadeImpl.java#L31
* https://github.com/devonfw/devon4j/blob/9b8f9461997bb5d8048a9fec42bec3625652eec3/modules/service/src/main/java/com/devonfw/module/service/common/impl/header/ServiceHeaderCustomizerCorrelationId.java#L42
* You can even have a look at the log-files of any devon4j app

For open-tracing there is already something in kafka:
* https://github.com/devonfw/devon4j/blob/bdd0c44c53c1ff91fb853e1aa5c12f245cb256c1/modules/kafka/src/main/java/com/devonfw/module/kafka/common/messaging/trace/impl/MessageTraceSupport.java

maihacke:
I did some more research on this, there are multiple solutions available. I propose that we incorporate the relativly new Opentracing API.

1. Related to this there is out-of-the-box support in Apache CXF: http://cxf.apache.org/docs/using-opentracing.html
2. Addtionally we need to investigate if we really need a fully featured opentracing implementation like Jaeger. I think a simplicstic approach which puts the tracing information into the logs / MDC would be suffient. Users of Devon4j should still be able to replace the opentracing implementation with something more complex, if they need the features.
[...]

PR #278, opened 2020-07-27
Draft PR of issue #260

ravicm2:


modules/service/pom.xml
```diff
@@ -47,6 +48,16 @@
       <artifactId>javax.annotation-api</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-api</artifactId>
+      <version>0.33.0</version>
```
maihacke, line 54: This version should go to the bom

modules/service/pom.xml
```diff
@@ -47,6 +48,16 @@
...
+    </dependency>
+    <dependency>
+      <groupId>io.jaegertracing</groupId>
+      <artifactId>jaeger-client</artifactId>
+      <version>0.32.0</version>
```
maihacke, line 59: This version should go to the bom

modules/service/src/main/java/com/devonfw/module/service/common/impl/header/ServiceHeaderCustomizerTraceSpanId.java
```diff
@@ -0,0 +1,107 @@
...
+    try (Scope scope = tracer.scopeManager().activate(span)) {
+
+      context.setHeader(LoggingConstants.TRACE_ID, span.context().toTraceId());
+      context.setHeader(LoggingConstants.SPAN_ID, span.context().toSpanId());
+
+    } catch (Exception ex) {
+      Tags.ERROR.set(span, true);
+      span.log(Map.of(Fields.EVENT, "error", Fields.ERROR_OBJECT, ex, Fields.MESSAGE, ex.getMessage()));
+    } finally {
+      span.finish();
```
maihacke, line 102: I do not get the point here? why is the span immediatly closed after setting the headers? Should the span be closed after the http request is made?

ravicm2: @maihacke  i thought a for each rest call, the request comes to this ServletHeaderContext class. so the setting the span to the headers and closing it. for the next call , new span is created , logged, set to the headers and closed.

Issue #269, opened 2020-05-27
Wiki: Link to Database Options is broken


phschimm:
The link under `Options > Databases` in the sidebar leads to a 404-page. It's supposed to link to:
https://github.com/devonfw/devonfw-guide/blob/master/general/db/guide-database.asciidoc

**EDIT:**
Actually not such an easy fix since (according to [.travis.yml](https://github.com/devonfw/devon4j/blob/develop/.travis.yml)) every `.asciidoc` extension gets removed while cloning to the `.wiki` repo. This is fine for relative links to files in the `documentation` directroy, but breaks external links. :confused: 
I wonder why such a complicated config is still used for this simple process. There's a "[Publish to GitHub Wiki](https://github.com/marketplace/actions/publish-to-github-wiki)" action, that could simplify this process massively.

hohwille:
@phschimm thanks for reporting this. Also thanks for your suggestions to improve and simplify.
IMHO the actual complexity and problems come from github itself as it renders AsciiDoc or Markup in generally totally inconsistent and different in each context (issue/PR comments, markup files in git repos, wiki pages, etc.). So while in one place the `.asiidoc` extension is required and in the other place it must not be present.
It is not that we want this to be complex - it is only that we want to have it working.

If the suggested "Publish to GitHub Wiki" action can really solve all these problems, I am very happy to replace our custom process with it. Have you already tested it (e.g. by forking this repo and configuring that github action for your fork)?

phschimm:
I wanted to test this and already ran into an inherent problem with the GitHub Actions system:

If you just use an action from the marketplace and its creator changes something, your whole pipeline could very well break ... This just happened to one of my personal repos using the "[Publish to GitHub Wiki](https://github.com/marketplace/actions/publish-to-github-wiki)" action. :unamused:

So it's probably better to have some quirks but full control, instead of quirks and no control. :sweat_smile:

hohwille:
The problem comes from this simple command that is copy&pasted to all repos:
https://github.com/devonfw/devon4j/blob/c03f577955d9e0cd98943db2687306b43b07df3d/.travis.yml#L41

Ideally we should have a `tool` like `docgen` that is used by our CI pipeline so we do not have to update all our `.travis.yml` files if thinks like this have to change.

sjimenez77:
[...]

PR #285, opened 2020-08-25
Wiki link to database options is broken #269

suvmanda:
Fixed Wiki: Link to Database Options is broken #269


documentation/guide-blazegraph.asciidoc
```diff
@@ -0,0 +1,13 @@
+:toc: macro
```
hohwille, line 1: My intention was not to duplicate the database guides to devon4j as they have intentionally been moved away from devon4j to the devonfw guide before in order to be available for entire devonfw cross-cutting (not Java specific).
With manually duplicating the content we will end up in a maintenance problem since changes then have to be done in both places what will surely not work out.
Actually the link you replaced in the _Sidebar is working perfectly correct. So I do not get the problem.
Seems I have to study the issue #269 or did you actually get what link exactly is broken?

hohwille (changes requested):

documentation/guide-blazegraph.asciidoc
suvmanda: I feel the link : https://github.com/devonfw/devonfw-guide/blob/master/general/db/guide-database is not taking the **.asciidoc** and the complete url should be : https://github.com/devonfw/devonfw-guide/blob/master/general/db/guide-database.asciidoc.

hohwille: Yep. Indeed the URLs (and also `link:` references) need to be complete including `.asciidoc` extension. Only in wikis the `link` references exclude the `.asciidoc` extension, but the wiki content is to be autogenerated by our CI pipelines and never maintained manually in `devonfw`.

hohwille:
Now that we understood the real source of the problem, we can IMHO close this PR. To fix #269 we need a new PR that fixes the `.travis.yml` file.

Issue #289, opened 2020-09-01
Upgrade travis dist to bionic


hohwille:
As suggested by @markusschuh it is about time to update the `dist` of our travis jobs since we are still on `trusty` from 2014. I suggest to update to bionic (more recent than default but not "latest bleeding edge" if you can say so for ubuntu/debian at all). If all works out, we should update all our `.travis.yml` files in devonfw accordingly.

hohwille:
works smooth. Other repos will follow...

PR #290, opened 2020-09-01
#289: update travis dist to bionic

hohwille:
Fix for #289


hohwille:
For this one line change in `.travis.yml` the successful build of travis seems sufficent so I skipped review and merged directly with admin permission. In the travis log I verified that `bionic` was actually used: all is working fine.

Issue #288, opened 2020-08-31
Update jackson to 2.11.2


hohwille:
As a devon4j user, I want to get new apps based on recent technology-stack so that I start a new project on recent libs rather than having known bugs and CVEs from the beginning leading to technical debt.

Current jackson version is `2.10.3`:
https://github.com/devonfw/devon4j/blob/b540a8e0dc26303d2e3c7ac6dcf16a8ce172d758/pom.xml#L35

Latest available version is `2.11.2`:
https://repo1.maven.org/maven2/com/fasterxml/jackson/jackson-bom/

We should also consider if latest `spring-boot` already ships with most recent jackson and we do not need to override anymore (see #287).

PR #292, opened 2020-09-01
#288: update jackson to 2.11.2

hohwille:
Implements #288


hohwille:
Tested isolated and in combination with PR #291.

Issue #295, opened 2020-09-08
Issue in running executable jar 


ssarmokadam:
### Expected behavior

As a user , I want to start newly created application using bootified jar with command java -jar jarname.

### Actual behavior

When executed this command it gives error "no main manifest attribute"

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  Create new devon4j application.
2. In  core module\target folder launch jar file with command java -jar jarfilename

### Affected version:

* OS: Windows

ssarmokadam:
Right way to test it is to run bootified war from server module. So this is an invalid issue.

hohwille:
Maybe we can take this issue as an opportunity to improve our documentation.
My suggested place would be to add a section here:
https://github.com/devonfw/devon4j/blob/develop/documentation/tutorial-newapp.asciidoc
Also feel free to add a small section `make jar not war` explaining how to change `server/pom.xml` such that a bootified war is created instead of bootified war for the `server` module.

PR #296, opened 2020-09-09
Fix for bootified jar

ssarmokadam:
Fix for issue https://github.com/devonfw/devon4j/issues/295 

After creating project with updated archetype we should be able to run jar with command java -jar jarname.


templates/server/src/main/resources/archetype-resources/core/pom.xml
```diff
@@ -252,6 +252,17 @@
           </excludes>
         </configuration>
       </plugin>
+      <plugin>
+		<groupId>org.springframework.boot</groupId>
```
hohwille, line 256: Sorry for being picky, but can you format this and fix the indentation.
If you are using latest devonfw-ide with eclipse all should be preconfigured and you only need to format (`[shift][ctrl][F]`).

hohwille (changes requested):
@ssarmokadam Thanks for testing and trying to fix. However, I assume we have a missunderstanding already in the "bug" issue. Can you double check? If I am right, we can close issue as invalid and close this PR. Otherwise I need to understand the context better.

ssarmokadam:
not required

Issue #286, opened 2020-08-28
Travis broken: The job exceeded the maximum log length, and has been terminated.


hohwille:
### Expected behavior

As a devon4j product owner, I want to have working and reliable CI so that I can monitor and garantee code health.

### Actual behavior

Travis-CI failes due to technical limitations:
```
The job exceeded the maximum log length, and has been terminated.
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. run travis on devon4j

### Related/Dependent Issues

PR #280 (where the error occurred the first time)

hohwille:
What can we do?
* switch to travis enterprise
* run maven quitely
* suppress more log statements by raising thresholds (https://github.com/devonfw/devon4j/blob/develop/modules/test/src/main/resources/logback.xml#L16)

maybeec:
Migrate to github actions?

suvmanda:
The build failed due to .travis.yml file "dist: trusty". when use "dist: bionic" the build is successful.

hohwille:
It was already indeterministic before and some builds succeeded while others failed.
If we are not experiencing the problem anymore with recent builds, it is fine to just close this issue.
I will observe some more builds before doing so.

suvmanda:
This is under observation. In future if not comes the problem with the builds ,we will close the issue.

hohwille:
problem did not occur anymore - hence closing.

PR #297, opened 2020-09-15
#286: Travis broken issue fixed

suvmanda:


hohwille:
So all this PR is showing that the build failed after adding `-q` and succeeded once after removing again.
The remaining changes of this PR is an appended whitespace, so IMHO you can already close the PR.
I am not yet convinced if we really understood and fixed the problem. To me it seems as the behaviour is indeterministic and sometimes the build failes with this error and sometimes not.

Issue #71, opened 2019-03-26
improve documentation on password encryption for strong encryption


maybeec:
I made use of jasypt password encryption in my property files.
One of my first step was to choose a strong cipher algorithm, so I was taking `PBEWITHHMACSHA512ANDAES_256` as this was the best provided by jasypt at this moment of time. Interestingly, running the application, you get login issues as of non-provided password.
During debugging I found out, that the password is decrypted to empty string for some reason. I haven't found the cause yet.

Anyhow, googling further, it seems, that jasypt is very instable in regards to AES encryption:
https://sourceforge.net/p/jasypt/bugs/32/

Are there alternatives right now? Do we wait until jasypt is released in a proper version?

maybeec:
also see ulisesbocchio/jasypt-spring-boot#124

hohwille:
This is only related to the documentation. Fixes need to be applied here:
https://github.com/devonfw/devon4j/blob/develop/documentation/guide-configuration.asciidoc#password-encryption

hohwille:
It is a pitty: We had this story almost done with PR #298 but as the author of the PR left without reworking the last review feedback and nobody took over, I have to move this issue to the next release :(

hohwille:
@maybeec I have merged #332. Can you check if that now matches your expectation and the new recommended algorithm is addressing your concerns of this issue so we can close it?

hohwille:
What still confuses me is why PR #332 documented the algorithm `PBEWITHHMACSHA512ANDAES_256` that you suggested but also added this config suggestion:
```
jasypt.encryptor.algorithm=PBEWITHMD5ANDTRIPLEDES
```

BTW: Shouls that algorithm better go to `application.yml` or to `application.properties`? If it goes to `application.properties` and this is our best practice - why dont we already ship this in the main `application.properties` of our archetype?

maybeec:
I am quite confused that we have both. Having both next to each other ´application.yml` and `application.properties` does not make sense to me.

Has encryption and running the application been tested with PBEWITHHMACSHA512ANDAES_256? just asking as I had some issues in the past to get it running.

hohwille:
> I am quite confused that we have both. Having both next to each other ´application.ymlandapplication.properties` does not make sense to me.

[...]

PR #298, opened 2020-09-24
rewritten doc about jasypt password encryption and decryption

ravicm2:


ravicm2:
https://github.com/devonfw/devon4j/issues/71

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
 Often you need to have passwords (for databases, third-party services, etc.) as part of your configuration. These are typically environment specific (see above). However, with DevOps and continuous-deployment you might be tempted to commit such configurations into your version-control (e.g. `git`). Doing that with plain text passwords is a severe problem especially for production systems. Never do that! Instead we offer some suggestions how to deal with sensible configurations:
 
 === Password Encryption
-A simple but reasonable approach is to configure the passwords encrypted with a master-password. The master-password should be a strong secret that is specific for each environment. It must never be committed to version-control.
-In order to support encrypted passwords in spring-boot `application.properties` all you need to do is to add https://github.com/ulisesbocchio/jasypt-spring-boot#jasypt-spring-boot[jasypt-spring-boot] as dependency in your `pom.xml`(please check for recent version):
-[source, xml]
+Let us have a look how to encrypt password using `jasypt-springboot-starter` and `jasypt-springboot`.
```
hohwille, line 159: should be `jasypt-spring-boot-starter`

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
...
-  <groupId>com.github.ulisesbocchio</groupId>
-  <artifactId>jasypt-spring-boot-starter</artifactId>
-  <version>1.17</version>
+    <groupId>com.github.ulisesbocchio</groupId>
+    <artifactId>jasypt-spring-boot-starter</artifactId>
+    <version>2.0.0</version>
 </dependency>
 ----
+
+Please check the latest version link:https://mvnrepository.com/artifact/com.github.ulisesbocchio/jasypt-spring-boot-starter[here] .
```
hohwille, line 175: nice. Thanks 👍

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
...
+
 This will smoothly integrate http://jasypt.org/[jasypt] into your https://projects.spring.io/spring-boot/[spring-boot] application. Read this https://wiki.jasig.org/display/CASUM/HOWTO+Use+Jasypt+to+encrypt+passwords+in+configuration+files[HOWTO] to learn how to encrypt and decrypt passwords using jasypt. Here is a simple example output of an encrypted password (of course you have to use strong passwords instead of `secret` and `postgres` - this is only an example):
+
 [source, bash]
 ----
-----ARGUMENTS-------------------
+ARGUMENTS
```
hohwille, line 181: just for clarification:
This was the raw output of jasypt CLI.
Has this been changed, as I produced invalid ASCIIDOC since the 5 dashes indicate the end of the source?
We could also simply change the 5 dashes opening and ending to 3 backticks instead.

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
...
 input: postgres
 password: secret
 
-----OUTPUT----------------------
+OUTPUT
 
 jd5ZREpBqxuN9ok0IhnXabgw7V3EoG2p
 ----
 
-The master-password can be configured on your target environment via the property `jasypt.encryptor.password`. As system properties given on the command-line are visible in the process list, we recommend to use an `config/application.yml` file only for this purpose (as we recommended to use `application.properties` for regular configs):
```
hohwille, line 182: why has this section been removed? It IMHO contains relevant information that is now missing.

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
...
+
+    @Value("${encrypted.property}")
+    private String property;
+
+    public String getProperty() {
+        return property;
+    }
+}
+----
+Using the above service class and setting the secret key which we used for encryption, we can easily retrieve the encrypted.property.
```
hohwille, line 277: This is an example that sounds more like a "tutorial to get started with jasypt" to me. IMHO we need a clear pattern/best practice how to do it properly.

documentation/guide-configuration.asciidoc
```diff
@@ -156,47 +156,189 @@ We recommend the following base layout for the hierarchical business configurati
...
+    config.setPassword("password");
+    config.setAlgorithm("PBEWithMD5AndDES");
+    config.setKeyObtentionIterations("1000");
+    config.setPoolSize("1");
+    config.setProviderName("SunJCE");
+    config.setSaltGeneratorClassName("org.jasypt.salt.RandomSaltGenerator");
+    config.setStringOutputType("base64");
+    encryptor.setConfig(config);
+    return encryptor;
+}
```
hohwille, line 300: So are these the parameters that make the encrpytion secure according to state of the art?
Did you align this with @maybeec or where did you get this code from?

hohwille (changes requested):
@ravicm2 thanks for all your investigation to take this further and for creating this PR. 👍 
IMHO this story is not so easy and I added some questions and review comments. Let us see what we can clarify easily and quickly. For all remaining aspects we might better do a call together...

documentation/guide-configuration.asciidoc
hohwille: should we also replace the example data to fit with the new encryption algorithm?

ravicm2: @hohwille  do you mean example provided here in the document or any other application already there for jasypt?

documentation/guide-configuration.asciidoc
ravicm2: I have added this at line number 327

documentation/guide-configuration.asciidoc
ravicm2: No i have not discussed regarding this with @maybeec but this is an example i learned from a tutorial.

ravicm2:
> @ravicm2 thanks for all your investigation to take this further and for creating this PR. 👍
> IMHO this story is not so easy and I added some questions and review comments. Let us see what we can clarify easily and quickly. For all remaining aspects we might better do a call together...

Sure @hohwille . May be tomorrow we can have a call regarding this. May be we can include @maybeec if he is free.

documentation/guide-configuration.asciidoc
hohwille: I was only talking about the outlook of jasypt CLI where I annotated this review comment. With example I am talking about the used data
```
input: postgres
password: secret
```
and especially the produced result that IMHO will change as the suggested security algorithm changed with this PR. Also a question would be if the CLI changes or a different Main-Class needs to be used for this now.

documentation/guide-configuration.asciidoc
hohwille: But you "removed" the info that the property is called `jasypt.encryptor.password`. I always welcome if people make our documentation more detailed/expressive, but we should avoid removing valuable info.

documentation/guide-configuration.asciidoc
maybeec: This config allows you to adapt the algorithm.
Anyhow the algorithm used here is the standard one which is quite insecure. Anyhow, at the time I tried making it more secure I was not able to get any other algorithm to work for any reason. Possibly this has been fixed meanwhile!?

hohwille:
This PR is IMHO replaced by PR #332.
I therefore close it.

PR #304, opened 2020-09-28
Fix macros shown on devonfw.com

jambulud:


hohwille (approved):
If this change fixes our website, I am fine with it. Seems that our tooling is not properly asciidoc conform.

PR #305, opened 2020-09-30
Updated tutorial for new app.

ssarmokadam:
Added section make jar not war and how to run app using created jar/war file.


ssarmokadam:
wrong files got committed

PR #302, opened 2020-09-24
#295: added guide for project/app structure

hohwille:
Improved documentation as suggested in #295

PR #306, opened 2020-09-30
improve documentation to launch app

ssarmokadam:
added section make jar not war. how to run app with bootified jar & war.

### DIFF
--- a/documentation/tutorial-newapp.asciidoc
+++ b/documentation/tutorial-newapp.asciidoc
@@ -69,20 +69,12 @@ The toplevel `pom.xml` of the generated project has the following features:
 * Maven plugins with desired versions and configuration
 * Profiles for link:guide-testing.asciidoc[test stages]
 
-== Make jar not war
+== How to run your app
 
-By default server module will bundle entire app as WAR file. To create bootified jar instead of war user needs to make below changes in server/pom.xml`
+=== Run app from IDE
 
-* In server/pom.xml update <packaging>war</packaging> to <packaging>jar</packaging> .
-* once pom.xml is updated execute maven build command such as mvn clean install.
-* Check server/target folder user should see ${projectname}-server-bootified.jar file created.
+To run your application from your favourite IDE, simply launch `SpringBootApp` as java application.
 
-== How to run project 
+=== Run app as bootified jar or war
 
-=== Run Project from IDE :
-
-To run project from ide user needs to launch `SpringBootApp` as java application.
-
-=== Run Project using bootified jar/war:
-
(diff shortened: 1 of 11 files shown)


hohwille:
@ssarmokadam Thanks for this PR.
Sorry, that I did not let you know but as I did not see this comming I have meanwhile done the same in PR #302 - we can now decide which text is nicer or which spot in the documentation is better or even combine both PRs to make the best out of it...

hohwille:
Your additions how to launch the app are also helpful.
IMHO we should mere #302 (review required) and after that slightly adopt this PR (link to the existing make jar not war section from #302, use backticks for classnames like `SpringBootApp`) and then merge also this PR.

documentation/tutorial-newapp.asciidoc
```diff
@@ -68,3 +68,22 @@ The toplevel `pom.xml` of the generated project has the following features:
...
+
+* In server/pom.xml update <packaging>war</packaging> to <packaging>jar</packaging> .
+* once pom.xml is updated execute maven build command such as mvn clean install.
+* Check server/target folder user should see ${projectname}-server-bootified.jar file created.
+
+== How to run project 
+
+=== Run Project from IDE :
+
+To run project from ide user needs to launch SpringBootApp.java as java class.
```
hohwille, line 84: ... to launch `SpringBootApp` as java application.

hohwille (changes requested):
I merged #302 and updated this PR. @ssarmokadam could you be so kind and update by linking to the new existing "make jar not war" section from #302 instead of repeating it here.

documentation/tutorial-newapp.asciidoc
```diff
@@ -68,3 +68,22 @@ The toplevel `pom.xml` of the generated project has the following features:
...
+
+=== Run Project using bootified jar/war:
+
+* To run project using bootified jar/war , navigate to server module target folder. In target folder user will get ${projectname}-server-bootified.war or ${projectname}-bootified.jar file.
```
hohwille, line 88: Also use backticks for all technical names here: `target` instead of just target.
Further it needs to be `${projectname}-server-bootified.*` in both cases (you omitted the `-server` in the JAR filename).

hohwille: actually also described here: https://github.com/devonfw/devon4j/blob/develop/documentation/guide-structure.asciidoc#make-jar-not-war

documentation/tutorial-newapp.asciidoc
ssarmokadam: Done

hohwille (approved):

Issue #283, opened 2020-08-19
Support for CompletableFuture in async service client


hohwille:
As discussed in PR #280 for issue #279 we should additionally consider supporting to return `CompletableFuture<R>` as a result of an async invocation rather than taking a callback:

```
CompletableFuture<R> call(R result);
```

PR #313, opened 2020-11-17
Support for CompletableFuture in async service client 

ssarmokadam:
Changes for https://github.com/devonfw/devon4j/issues/283


modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java
```diff
@@ -117,4 +117,15 @@ protected Object handleUnsupportedBody(Object body) {
    */
   protected abstract HttpRequest createRequest(ServiceClientInvocation<S> invocation);
 
+  // START
```
hohwille, line 120: This does not seem helpful to me, can you remove this?

modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java
```diff
@@ -117,4 +117,15 @@ protected Object handleUnsupportedBody(Object body) {
...
+  @Override
+  protected <R> CompletableFuture<R> getCompletableFuture(ServiceClientInvocation<S> invocation) {
+
+    long startTime = System.nanoTime();
```
hohwille, line 124: this variable is not used anywhere, can you remove this line?

modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java
```diff
@@ -117,4 +117,15 @@ protected Object handleUnsupportedBody(Object body) {
...
+    HttpRequest request = createRequest(invocation);
+    return (CompletableFuture<R>) this.client.getHttpClient().sendAsync(request, BodyHandlers.ofString())
+        .thenApply(HttpResponse::body);
```
hohwille, line 127: When adding new features to devon4j we should always consider testing and documenting them.
You can not cast from `CompletableFuture<String>` to `CompletableFuture<R>` so IMHO this code can not work properly (except `R` happens to be `String`). There is the existing logic already present from the other `call` methods that  transforms the `HttpResponse` to the result (`R`) by unmarshalling from JSON. You should reuse this and transform the `CompletableFuture` based on that logic.

modules/service/src/main/java/com/devonfw/module/service/common/api/client/AsyncServiceClient.java
```diff
@@ -64,4 +65,12 @@
    */
   void callVoid(Runnable serviceInvoker, Consumer<Void> resultHandler);
 
+  // START
```
hohwille, line 68: can you remove this line?

modules/service/src/main/java/com/devonfw/module/service/common/api/client/AsyncServiceClient.java
```diff
@@ -64,4 +65,12 @@
...
+  /**
+   * @param <R> type of the result of the service operation to call.
+   * @param result
+   * @return
```
hohwille, line 72: Please always write complete JavaDoc esp. for interfaces and major APIs.

modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
```diff
@@ -120,4 +121,26 @@ private static int getLength(Object[] parameters) {
     return parameters.length;
   }
 
+  // START
```
hohwille, line 124: please remove this line.

modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
```diff
@@ -120,4 +121,26 @@ private static int getLength(Object[] parameters) {
...
+    return future;
+
+  }
+
+  /**
+   * @param <R>
+   * @param invocation
+   * @return
+   */
+  protected abstract <R> CompletableFuture<R> getCompletableFuture(ServiceClientInvocation<S> invocation);
```
hohwille, line 144: IMHO `getCompletableFuture` is somewhat missleading as this is not a "getter". Why dont we just call this method `call` as well?

modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
```diff
@@ -120,4 +121,26 @@ private static int getLength(Object[] parameters) {
...
+      this.errorHandler.accept(t);
+    }
+    return future;
+
+  }
+
+  /**
+   * @param <R>
+   * @param invocation
+   * @return
```
hohwille, line 142: Please complete JavaDoc.

hohwille (changes requested):
@ssarmokadam thanks for this PR.
Could you add/extend JUnits to test this new feature? In the corresponding starter you can find existing JUnit that tests the async call via callback. If you test this including JSON marshalling we can easily verify if I am right that your code does not work as commented in the review.
Can you also extend the documentation accordingly:
https://github.com/devonfw/devon4j/blob/develop/documentation/guide-service-client.asciidoc#simple-usage
Please see the review comments for further details.

hohwille (changes requested):
Thanks for taking this further. However the idea of `Future`s is that the computation is asynchronous and non-blocking. Your code is using `future.get()` in the doCall method what breaks this asnynchronous processing.
If it helps you may read some documentation about this:
https://www.callicoder.com/java-8-completablefuture-tutorial/

hohwille:
Any chance we can complete this story and include it into the release?

modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java
```diff
@@ -117,4 +119,39 @@ protected Object handleUnsupportedBody(Object body) {
...
+        return (R) result;
+      }
+    } catch (Throwable t) {
+      this.errorHandler.accept(t);
+      error = t;
+    } finally {
+      ServiceClientPerformanceLogger.log(startTime, service, response.statusCode(), error);
+    }
+    return null;
+  }
```
hohwille, line 155: This method is not new. Is that a bug of github that it is showing this as newly added in the diffs?
Here is the existing method before this PR:
https://github.com/devonfw/devon4j/blob/3268e8e563ec1da885ad07467e41aa8cbfd904c5/modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java#L57

@ssarmokadam There is IMHO nothing wrong from your end, I am just confused by what github is showing me to review...

hohwille (approved):
@ssarmokadam thanks for figuring out the proper solution to deal with futures. This was surely not an easy story without further assistance. Thanks for solving this nicely without further attention 👍 
This is now ready to be merged.

PR #312, opened 2020-11-17
#307: Csrf starter

sujith-mn:

PR #314, opened 2020-11-27
template changes for CSRF

sujith-mn:


hohwille:
IMHO we can close this PR that is replaced by #312

Issue #307, opened 2020-10-08
Fix CSRF protection support


jdiazgon:
### Description

Current devon4ng application implements a method to retrieve CSRF token from the backend:

```TypeScript
  getCsrf(): string {
    return this.serverPath + 'security/v1/csrftoken';
  }
```
From [this file](https://github.com/devonfw/devon4ng-application-template/blob/develop/src/app/core/shared/business-operations.service.ts#L17).

UPDATE: new URL path is `csrf/v1/token`.

This worked like a charm months ago. But it seems the 24th of April, this service `csrftoken` was completely removed from devon4j. I think the related PR is #244 

As I can see in [this file](https://github.com/devonfw/devon4j/blob/9fcfd8f48ca229578214c358678743f327c93e55/templates/server/src/main/resources/archetype-resources/api/src/main/java/__packageInPathFormat__/general/service/api/rest/SecurityRestService.java), the service was defined:

```Java
  @GET
  @Path("/csrftoken/")
  CsrfToken getCsrfToken(@Context HttpServletRequest request, @Context HttpServletResponse response);
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Download https://github.com/devonfw/devon4ng-application-template/
2. Set environment.ts file to use `csrf`
3. Try to login and you will face 404, statusText: "OK", url: "http://localhost:8081/services/rest/security/v1/csrftoken 

### Comments/Hints:

Has devon4j completely disabled CSRF support? Do we have now to use JWT for the front-end?

hohwille:
First of all we have to clarify what the actual bug is here.
IMHO you are talking about the `SecurityRestService`:
https://github.com/devonfw/devon4j/blob/4e00cf2d746b1038afcc65d0830bc3ba09c4f213/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/service/impl/rest/SecurityRestServiceImpl.java

This has been removed from our app template.
However, IMHO this is NOT a bug but has been done on purpose as people demanded to app template to be minimalistic and they want `opt-in` rather than `opt-out`.
So first we have to clarify directions before doing any change.
I can not find the commit that deleted this `SecurityRestServiceImpl` but I assume it is related to issue #171.

What confuses me also is that we have removed this service but still have `CSRF` enabled by default:
[...]

PR #316, opened 2020-12-01
#307: design proposal for modular security

hohwille:
Design proposal for #307 to be merged with PR #312

### DIFF
--- a/boms/minimal/pom.xml
+++ b/boms/minimal/pom.xml
@@ -77,6 +77,11 @@
       </dependency>
       <dependency>
         <groupId>com.devonfw.java.modules</groupId>
+        <artifactId>devon4j-security-csrf</artifactId>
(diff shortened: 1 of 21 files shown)


hohwille:
As an improvement we could change CsrfRestServiceImpl to transform the CsrfToken instance manually into CsrfTokenImpl that will be moved to csrf module. Then we can remove it from archetype including the jackson ObjectMapperFactory config. As a result CSRF would be entirely modular without further side-effects.

Further, we IMHO need to adopt devon4ng as the REST path to get CSRF token changed. Is someone take care of all these aspects so all fits together in the end? Any ideas how to deal with existing devon4j apps upgrading to new devon4j and devon4ng?

hohwille:
Closing as this was only input to support #312

PR #322, opened 2020-12-14
devon4j application failed to start

sujith-mn:
Hotfix for #312


hohwille (approved):
@sujith-mn OK. I see. So without this dependency, the `webSecurityConfigurer.configure(http)` will not work or the `webSecurityConfigurer` can not even be injected.
Thanks for testing and fixing! 👍 
Can someone explain why our integration tests of our archetype do not reveal such problem? IMHO every spring-boot integration test must fail without this fix then. But we have some of such tests in our archetype that should run during the integration test?!?

hohwille:
That should also be covered during tests:
https://github.com/devonfw/devon4j/blob/develop/templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/impl/config/TestWebSecurityConfig.java#L33

hohwille:
I checked the archetype. Fun-fact is that there is no springboottest left. We only have `RestServiceTest` which is abstract and acts as base-class for REST tests.

Issue #324, opened 2020-12-15
Nullpointerexception in CompletableFuture AsyncServiceClient


sujith-mn:
Got the below error while calling the remote service asynchronously.

2020-12-15 21:05:42.129  INFO 11620 --- [onPool-worker-3] m.s.c.b.c.ServiceClientPerformanceLogger : Invoking service com.example.domain.test1.employeemanagement.service.impl.rest.AsyncEmployeeService#getEmployee[http://localhost:8080/services/rest/employeemanagement/v1/employee/1] took PT47.4093297S (47409329700ns) and succeded with status 302.
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at com.example.domain.test1.employeemanagement.service.impl.rest.EmployeemanagementRestServiceImpl.getEmployee(EmployeemanagementRestServiceImpl.java:49)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.apache.cxf.service.invoker.AbstractInvoker.performInvocation(AbstractInvoker.java:179)
	at org.apache.cxf.service.invoker.AbstractInvoker.invoke(AbstractInvoker.java:96)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:201)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:104)
	at org.apache.cxf.interceptor.ServiceInvokerInterceptor$1.run(ServiceInvokerInterceptor.java:59)
	at org.apache.cxf.interceptor.ServiceInvokerInterceptor.handleMessage(ServiceInvokerInterceptor.java:96)
	at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:308)
[...]

PR #325, opened 2020-12-15
#324: fix of NPE in error handling

hohwille:
Fix for NPE in error handling identified in bug issue #324.

### DIFF
--- a/modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
+++ b/modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
@@ -30,6 +30,9 @@ public abstract class AbstractAsyncServiceClient<S> implements AsyncServiceClien
   /** The {@link #setErrorHandler(Consumer)} */
   protected Consumer<Throwable> errorHandler;
 
+  /** The most recent invocation. */
+  private ServiceClientInvocation<S> invocation;
+
   /**
    * The constructor.
    *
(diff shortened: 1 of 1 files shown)


modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
```diff
@@ -30,6 +30,9 @@
   /** The {@link #setErrorHandler(Consumer)} */
   protected Consumer<Throwable> errorHandler;
 
+  /** The most recent invocation. */
+  private ServiceClientInvocation<S> invocation;
```
maybeec, line 34: Is this thread safe? I am not deeply in this topic but I would guess that the abstract client class is basically representing one instance of a service interface.
If so, this would not be thread safe I think as two request would interfere each other

maybeec (changes requested):

modules/service/src/main/java/com/devonfw/module/service/common/base/client/async/AbstractAsyncServiceClient.java
hohwille: @maybeec thanks for your review and looking at things like this with a sharp eye 👍 

However, a service client is to be created per use and not be shared accross threads. Only `ServiceClientFactory` is thread-safe. Already before this change, 
Already before this change this class contains a `stub` that is not thread-safe.

maybeec: ok great, thanks for clarification!

maybeec (approved):

Issue #327, opened 2020-12-17
CompletableFuture returns `null` if service failed instead of raising exception


hohwille:
We have implemented issue #283 with PR #313 to have support for getting async service client result as `CompletableFuture`.
If for some reason the service invocation fails because of an error on the server in the backend providing the service, I observe the following behaviour:
* The backend responses with the error and faulty status code (in my case `500` Internal Server Error)
* The code to call asynchronously providing `CompletableFuture` transforms the result body: https://github.com/devonfw/devon4j/blob/7b1d0ecb100900abe952f84297e04e64547e405d/modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractAsyncServiceHttpClient.java#L129
* However, it does not check the status code and the body of the `500` response is empty and therefore `null` is returned.
* As a result the user of the API can not distinguish between proper service invocation without content (`204`) and faulty results (`4xx` or `5xx`).

hohwille:
Looking deeper `handleResponse` is properly handling the error but sends it to the `errorHandler`.
The existing code was designed to work with callbacks:
* the callback `successHandler` is used to receive the successful result
* the callback `errorHandler` is used to receive errors as `Throwable`

If an error occurs, only `errorHandler` should be invoked and the `successHandler` should not be invoked.
Now, with `CompletableFuture` this design does not directly make sense and both the error and the success has to be reported via the `CompletableFuture` to tell the invoker what happened asynchronously whilst invoking the service.

PR #328, opened 2020-12-17
#327: fix error handling for completable future

hohwille:
* Fixed bug #327 - throw exception from CompletableFuture
* extends JUnits to cover bug #324


maybeec (approved):

hohwille:
We could still invoke the `errorHandler` in before throwing the exception in the `CompletableFuture`.
Any thoughts on this design from others?
Currently this is not the case and I updated the JavaDoc to explain the behaviour. We could change this now if we think this would be better.

ssarmokadam (approved):

Issue #329, opened 2020-12-22
Create spring-boot starter for CORS


sujith-mn:
As a backend developer, I want to use CORS filter as a starter so that I can add it as a maven dependency whenever required.

PR #330, opened 2020-12-22
#329: Added CORS starter

sujith-mn:
Implementation of issue #329


hohwille (changes requested):
@sujith-mn thanks for this PR. I like your idea to follow the starter concept we introduced with `CSRF` also for `CORS`. This makes perfectly sense to me and your PR in general looks good and going in the right direction. 👍 

However, looking at the details your code seems incomplete and not ready to be merged. Please have a look at my review comments and rework.
It would also be great to have a JUnit proving that the filter and its actual configuration is really applied.

modules/service/src/main/java/com/devonfw/module/service/common/impl/CorsConfigProperties.java
```diff
@@ -0,0 +1,83 @@
...
+  private boolean allowCredentials;
+
+  private String allowedOrigin;
+
+  private String allowedHeader;
+
+  private String allowedMethods;
+
+  /**
+   * @return allowCredentials
```
hohwille, line 20: I actually do not know myself what it does. So as you can see JavaDoc would really help.
Also add fully-qualified `@see` tags to all these getters e.g.
```
@see org.springframework.web.cors.CorsConfiguration#isAllowCredentials()
```

starters/starter-security-cors/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+  <version>${devon4j.version}</version>
+  <packaging>jar</packaging>
+  <name>${project.artifactId}</name>
+  <description>Spring-boot starter for CORS protection.</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.devonfw.java.starters</groupId>
+      <artifactId>devon4j-starter-security</artifactId>
+       <version>2020.12.001-SNAPSHOT</version>
```
hohwille, line 21: never provide versions here, especially never hardcoded SNAPSHOTs.

templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/service/impl/config/BaseWebSecurityConfig.java
```diff
@@ -16,8 +16,6 @@
 import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.web.cors.CorsConfiguration;
```
hohwille, line 18: Also removed this dead import. You should create the archetype locally and import into Eclipse.
When you get warnings this means you need to rework :)

hohwille (changes requested):
@sujith-mn thanks for your great improvements. You nicely applied the `*ConfigProperties` pattern of `devon4j` here. Still there are some review comments to rework but the finishing line is close. 👍

starters/starter-security-cors/pom.xml
sujith-mn: This was done for my local testing.. accidently pushed in.. will be removed.

modules/security-cors/src/main/java/com/devonfw/module/security/cors/common/impl/CorsConfigProperties.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link ConfigurationProperties} to configure Cross-Origin Resource Sharing (CORS).
+ *
+ * @since 2020.12.002
+ */
+@Configuration
+@ConfigurationProperties(prefix = "security.cors")
+public class CorsConfigProperties {
```
hohwille, line 16: Lets change this entire class to this:
```
@Configuration
@ConfigurationProperties(prefix = "security.cors")
public class CorsConfigProperties extends CorsConfiguration {
}
```
This will give us all config options from springs `CorsConfiguration` and even future ones added in spring without changing our code.

modules/security-cors/src/main/java/com/devonfw/module/security/cors/common/impl/WebSecurityConfigurerAspectCors.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  private CorsFilter getCorsFilter() {
+
+    UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
+    CorsConfiguration config = new CorsConfiguration();
+    config.setAllowCredentials(this.corsConfigProperties.isAllowCredentials());
+    config.setAllowedOrigins(this.corsConfigProperties.getAllowedOrigins());
+    config.setAllowedHeaders(this.corsConfigProperties.getAllowedHeaders());
+    config.setAllowedMethods(this.corsConfigProperties.getAllowedMethods());
```
hohwille, line 37: All this mapping will not be needed anymore with my suggested improvement as you can use `this.corsConfigProperties` and directly pass it to `source`.

modules/security-cors/src/main/java/com/devonfw/module/security/cors/common/impl/WebSecurityConfigurerAspectCors.java
```diff
@@ -0,0 +1,41 @@
...
+    config.addAllowedOrigin("*");
+    config.addAllowedHeader("*");
+    config.addAllowedMethod("OPTIONS");
+    config.addAllowedMethod("HEAD");
+    config.addAllowedMethod("GET");
+    config.addAllowedMethod("PUT");
+    config.addAllowedMethod("POST");
+    config.addAllowedMethod("DELETE");
+    config.addAllowedMethod("PATCH");
+    source.registerCorsConfiguration("/**", config);
```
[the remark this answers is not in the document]
hohwille: This review comment seems still unresolved.

modules/service/src/main/java/com/devonfw/module/service/common/impl/CorsConfigProperties.java
```diff
@@ -0,0 +1,83 @@
+package com.devonfw.module.service.common.impl;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ *
+ *
+ */
+@ConfigurationProperties
+public class CorsConfigProperties {
```
hohwille, line 10: I see `CorsConfigProperties` twice here (the other on in `com.devonfw.module.security.cors.common.impl`). Is this one used after all or could it be deleted? IMHO this one is obsolete and should be deleted.

hohwille (changes requested):
Thanks for your improvements 👍 
I have a suggestion for a great simplification, can you have a look and test if that works as suggested?

modules/security-cors/src/main/java/com/devonfw/module/security/cors/common/impl/WebSecurityConfigurerAspectCors.java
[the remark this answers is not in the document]
hohwille: So IMHO we still have to address this issue here that was missed out so far.
My suggestion would be the following:
```
@ConfigurationProperties(prefix = "security.cors")
public class CorsConfigProperties {
   private final CorsConfiguration spring = new CorsConfiguration();
   private String pathPattern = "/**";

   /**
    * @return the {@link CorsConfiguration} from spring-web.
    */
  public CorsConfiguration getSpring() {
    return this.spring;
  }

  /**
   * @return the ant-style pattern for the URL paths where to apply CORS. Use "/**" to match all URL paths.
   * @see org.springframework.web.cors.UrlBasedCorsConfigurationSource#registerCorsConfiguration(String, CorsConfiguration)
   */
  public String getPathPattern() {
    return this.pathPattern;
  }

  /**
   * @param pathPattern the new value of {@link #getPathPattern()}.
   */
  public void setPathPattern(String pathPattern) {
    this.pathPattern = pathPattern;
  }
}
```

And then simply do:
```
  private CorsFilter getCorsFilter() {

    UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
    source.registerCorsConfiguration(this.corsConfigProperties.getPathPattern, this.corsConfigProperties.getSpring());
    return new CorsFilter(source);
  }
```

Also update the documentation to:
```
security.cors.spring.allowCredentials=true
security.cors.spring.allowedOrigins=*
security.cors.spring.allowedHeaders=*
security.cors.spring.allowedMethods=OPTIONS,HEAD,GET,PUT,POST,DELETE,PATCH
security.cors.pathPattern=/**
```

hohwille (changes requested):
One last review comment was missed out. Please check my suggestion.

modules/security-cors/src/main/java/com/devonfw/module/security/cors/common/impl/WebSecurityConfigurerAspectCors.java
```diff
@@ -12,7 +11,7 @@
 /**
  * Implementation of {@link WebSecurityConfigurerAspect} for CORS protection.
  *
- * @since 2020.12.001
+ * @since 2020.12.002
```
hohwille, line 14: That was my mistake. Excellent catch! Thanks. 👍

hohwille (approved):
Ready for merge. Thanks @sujith-mn for all your updates to make this PR perfect 👍

PR #332, opened 2021-01-05
#298 added configurable algorithm.

sujith-mn:


documentation/guide-configuration.asciidoc
```diff
@@ -167,14 +167,15 @@ In order to support encrypted passwords in spring-boot `application.properties`
 </dependency>
 ----
 This will smoothly integrate http://jasypt.org/[jasypt] into your https://projects.spring.io/spring-boot/[spring-boot] application. Read this https://apereo.atlassian.net/wiki/spaces/CASUM/pages/103261428/HOWTO+Use+Jasypt+to+encrypt+passwords+in+configuration+files[HOWTO] to learn how to encrypt and decrypt passwords using jasypt. Here is a simple example output of an encrypted password (of course you have to use strong passwords instead of `secret` and `postgres` - this is only an example):
+
 [source, bash]
 ----
-----ARGUMENTS-------------------
+----- ARGUMENTS ```
```
hohwille, line 173: is this valid AsciiDoc?
Either
```
[source, «type»]
---
```
or
```
\```
```
will start a code block.

If you did rerun jasypt with more recent algorithm you are more than welcome to replace this output examples here. However, I do not understand these changes here...

documentation/guide-configuration.asciidoc
```diff
@@ -196,7 +197,22 @@ To prevent jasypt to throw an exception in dev or test scenarios simply put this
...
+[source, java]
+----
+StandardPBEStringEncryptor encryptor = new StandardPBEStringEncryptor();
+String privateData = "secret-password-to-encrypt";
+encryptor.setPassword("master-password");
+encryptor.setAlgorithm("PBEWithMD5AndTripleDES");
+String encryptedText = encryptor.encrypt(privateData);
+----
+
+Other configurable properties can be found https://github.com/ulisesbocchio/jasypt-spring-boot#password-based-encryption-configuration[here]
```
hohwille, line 213: Very nice improvement. Still it is unclear what to do? Which algorithm should we use now?

hohwille (changes requested):
For #71 my expecation would be:
* we understand why the default algorithm is flawed and should be discouraged - if noone can clarify this and #71 is invalid, we should simply cose it. Otherwise we need to take action with the next points.
* we understand what algorithm should be our best-practice
* we practically proof the usage with the new algorithm on a running example application
* we change this documentation with concrete steps so all steps and example configs and outputs match to the new algorithm

hohwille (approved):
@sujith-mn thanks for your updates. I am approving. However, I still have one concern:
If we override the encryption algorithm manually in the jasypt CLI we IMHO also have to configure that in springs `application.properties`, right?
Did you test this and ensure that the password can be decrypted by the devon4j template application following this guide?

Otherwise if the new version of Jasypt has changed the encryption algorithm, do we still need to provide it in the jasypt CLI?

Appart from this last point this is ready for merge.

sujith-mn:
> @sujith-mn thanks for your updates. I am approving. However, I still have one concern:
> If we override the encryption algorithm manually in the jasypt CLI we IMHO also have to configure that in springs `application.properties`, right?
Yes, if other than default algorithm required, it needs to be configured.
> Did you test this and ensure that the password can be decrypted by the devon4j template application following this guide?
Yes
> Otherwise if the new version of Jasypt has changed the encryption algorithm, do we still need to provide it in the jasypt CLI?
If we do not provide the algorithm in CLI jasypt considers the default algorithm.

> Appart from this last point this is ready for merge.

Issue #164, opened 2019-10-27
Flattened pom for core project invalid


maybeec:
### Expected behavior

As an architect, I want package all modules of my software so that I can individually build and test each of the maven modules separately.

### Actual behavior

Having the core maven module deployed, a dependent software module build like the batch module states, that the deployed pom.xml of the core module is not valid as of dependency version not set for

```xml
        <dependency>
          <groupId>org.springframework.boot</groupId>
          <artifactId>spring-boot-starter-tomcat</artifactId>
        </dependency>
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. create new project by archetype
2. run mvn clean install on the entire multi-module project
3. go to batch module folder
4. run mvn clean test

### Related/Dependent Issues
mojohaus/flatten-maven-plugin#70

hohwille:
@maybeec Thanks for the feedback and also the link to the flatten issue you found.
Seems I do have some homework here. That is the drawback when you are starting many OSS projects trying to make the world better, you earn lots of maintenance effort. As I began flatten-maven-plugin more as a kind of PoC to show the maven community how a better maven should look like in the future, I might need to find another maintainer for this on the long run. However, for this issue I will try my best to find a fix and ship a new version. However, I can not primise this before end of November.

hohwille:
However, if I read your report carefully I come to the impression that we are creating an entirely invalid POM here. That does not only mean that the local build of the batch module is somehow problematic but the entire "release" of that module is kind of invalid and if some other project would import that dependency it would also not work.
I can not say how much of a blocker that is, but it sounds like a rather bad bug and not just a nice to have issue.

hohwille:
Via OSSRH SNAPSHOT deployment we can at least verify our POMs of devon4j now:
https://oss.sonatype.org/content/repositories/snapshots/com/devonfw/java/modules/devon4j-basic/3.3.0-SNAPSHOT/devon4j-basic-3.3.0-20200306.092248-3.pom

hohwille:
Can someone shed some light into this issue?
I created a devon4j project from the latest archetype from develop branch.
There is not flattening happening at all. Did we disable flatten plugin in archetype already as a quickfix for this issue long time ago?
In the toplevel POM of the archetype flatten plugin is configured in pluginManagement but the plugin itself is not used in build->plugins so the entire issue is invalid.
Can someone confirm before I close this issue?

hohwille:
[...]

PR #335, opened 2021-01-25
#164 temporary workaround to fix maven-flatten-bug

maybeec:
temporary fix for #164 as of mojohaus/flatten-maven-plugin#70


hohwille (approved):
@maybeec thanks for this PR 👍 
Sorry, it seems I missed out this PR for a while and just found it.
Looks like a reasonable workaround and quickfix and nice that you put the comment on top so we can later remove it when we can some day update flatten plugin.

PR #337, opened 2021-02-01
updated T-Architecture

hohwille:
Updated T-Architecture:
* migrated from Visio to draw.io
* fixed some typos (client components)
* fixed and improved links
* added DBs


maihacke:
In my opinion it would also be ok that UI-Component A1 uses Component A2 and A3 or even B.
In generel I often saw that the components in UI has to be cut different from the backend, e.g. because of many dialogs combining data from different components or even services. But that is a more general discussion, I think as an first step for getting away from visio it is ok.

maihacke (approved):

hohwille:
@maihacke thanks for your review and approval.

> In my opinion it would also be ok that UI-Component A1 uses Component A2 and A3 or even B.

We already have a case where A2 UI calls REST service from B. Aint that sufficient to express what you are thinking about?
Also, next we should rework this diagram here:
https://github.com/devonfw/sonar-devon4j-plugin/#architecture-rules

I will already merge this now and we can consider creating additional issues or PRs to improve.

maihacke:
I missed the (dotted) line from UI-A2 to B. That's exactly what I meant.

PR #340, opened 2021-02-05
Feature/liquibase configuration

dixyushi:


hohwille:
FYI: I typically do not start reviews before PRs are up-to-date with `master` and CI is succeeding.
I guess you have seen already in travis logs that we have integrated deeper with flyway so you need to add some switches there:
```
[INFO] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:testCompile (default-testCompile) on project app-with-batch-batch: Compilation failure: Compilation failure: 

[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/batch/project/app-with-batch/batch/src/test/java/it/pkg/general/batch/base/test/SpringBatchIntegrationTest.java:[5,25] package org.flywaydb.core does not exist

[INFO] [ERROR] /home/travis/build/devonfw/devon4j/templates/server/target/test-classes/projects/batch/project/app-with-batch/batch/src/test/java/it/pkg/general/batch/base/test/SpringBatchIntegrationTest.java:[24,11] cannot find symbol

[INFO] [ERROR]   symbol:   class Flyway

[INFO] [ERROR]   location: class it.pkg.general.batch.base.test.SpringBatchIntegrationTest
```

hohwille:
@dixyushi can you fix the merge conflicts and bring this PR to a "ready for review" state?

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/clean/TestCleanerPluginLiquibase.java
```diff
@@ -0,0 +1,45 @@
+package com.devonfw.module.test.common.base.clean;
+
+import javax.inject.Inject;
+import liquibase.Liquibase;
+
+/**
+ * @author ayudixit
+ *
+ */
```
sujith-mn, line 9: remove the author tag.

templates/server/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -13,6 +13,10 @@
       <defaultValue>h2</defaultValue>
       <!-- <validationRegex>h2|postgresql|mysql|mariadb|oracle|hana|db2</validationRegex> -->
     </requiredProperty>
+    <requiredProperty key="dbMigration">  
+      <defaultValue>flyway</defaultValue> 
+      <!-- <validationRegex>flyway|liquibase</validationRegex> -->  
+    </requiredProperty>
```
sujith-mn, line 19: remove the commented code.

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,73 @@
...
+import org.springframework.batch.core.launch.JobLauncher;
+import org.springframework.batch.test.JobLauncherTestUtils;
+import ${package}.general.common.base.test.TestUtil;
+import com.devonfw.module.test.common.base.ComponentTest;
+#if ($dbMigration == 'liquibase')  
+  import liquibase.Liquibase;
+#else 
+ import org.flywaydb.core.Flyway;
+#end
+
```
sujith-mn, line 16: make the alignment proper 
#if should start from 0th column, and import should align with the above imports.

sujith-mn (changes requested):

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/clean/TestCleanerPluginLiquibase.java
dixyushi: Removed and changes done.

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
dixyushi: Corrected the alignment and changed the code.

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/clean/TestCleanerPluginLiquibase.java
```diff
@@ -0,0 +1,48 @@
...
+    this.liquibase = liquibase;
+  }
+
+  @Override
+  public void cleanup() {
+
+    try {
+      this.liquibase.dropAll();
+    } catch (Exception e) {
+      e.printStackTrace();
```
hohwille, line 43: never use `printStackTrace()` in productive code. IMHO better than logging is to rethrow as runtime exception and let the test fail instead of silently ignoring a severe error.

pom.xml
```diff
@@ -1,67 +1,73 @@
...
-        <plugin>
-          <groupId>org.springframework.boot</groupId>
-          <artifactId>spring-boot-maven-plugin</artifactId>
-          <version>${spring.boot.version}</version>
-        </plugin>
-      </plugins>
-    </pluginManagement>
-  </build>
-</project>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hohwille, line 1: what has been changed here?
The diff shows every line has changed. EOL stlye changed between unix and linux?
The only content change I can see here is that you added `standalone="no"`. What was the reason for that change? I can not see any...

templates/pom.xml
```diff
@@ -1,19 +1,19 @@
...
-  <packaging>pom</packaging>
-  <name>${project.artifactId}</name>
-  <description>Templates (maven archetypes) of devon4j.</description>
-
-  <modules>
-    <module>server</module>
-  </modules>
-
-</project>
+<?xml version="1.0" encoding="UTF-8"?>
```
hohwille, line 1: same here? No change at all?

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,73 @@
...
-  public JobLauncherTestUtils getJobLauncherTestUtils(Job job) {
-
-    JobLauncherTestUtils jobLauncherTestUtils = new JobLauncherTestUtils();
-    jobLauncherTestUtils.setJob(job);
-    jobLauncherTestUtils.setJobLauncher(this.jobLauncher);
-
-    return jobLauncherTestUtils;
-  }
-}
+package ${package}.general.batch.base.test;
```
hohwille, line 1: again the diff is not readable. IMHO you have a general problem with your setup of git and are changing all EOLs accidentally. Can you please change and fix this?

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,73 @@
...
+  @Inject
+  private JobLauncher jobLauncher;
+
+  #if ($dbMigration == 'liquibase')  
+    @Inject
+    private Liquibase liquibase;
+  #else 
+    @Inject
+    private Flyway flyway;
+  #end
```
hohwille, line 31: indendation with velocity is tricky: every whitespace will be written to the output. That is "  " before the `#if` as well as "    " before `@Inject` resulting in 6 spaces before annotation and 4 spaces before private field whereas we should only have two spaces each.
BTW: I am wondering why we do not use the `TestCleanerPlugin` in this test to abstract from liquibase of flyway instead...

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,73 @@
...
+  
+  #if ($dbMigration == 'liquibase')  
+    @Override
+    protected void doSetUp() {
+      super.doSetUp();
+      try {
+        this.liquibase.dropAll();
+      } catch (Exception e) {
+        e.printStackTrace();
```
hohwille, line 40: see `TestCleanerPluginLiquibase`. IMHO we can avoid all this redundancy and velocity macro complexity by injecting and using `TestCleanerPlugin`.

templates/server/src/main/resources/archetype-resources/core/pom.xml
```diff
@@ -167,11 +167,20 @@
...
-    <!-- Flyway for DB Schema version management and migration -->
+    <!-- DB Schema version management and migration -->
+#if ($dbMigration == 'flyway')
     <dependency>
       <groupId>org.flywaydb</groupId>
       <artifactId>flyway-core</artifactId>
     </dependency>
+#end
+
+#if ($dbMigration == 'liquibase')
```
hohwille, line 178: use `#elseif` instead?

templates/server/src/main/resources/archetype-resources/core/pom.xml
```diff
@@ -254,7 +263,19 @@
           </excludes>
         </configuration>
       </plugin>
+      
+      #if($dbMigration == 'liquibase')
```
hohwille, line 267: can you make the indentdation consistent and also check the indendation of the generated result?

templates/server/src/main/resources/archetype-resources/core/src/main/resources/db/changelog/changelog-master.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hohwille, line 1: this file is only needed for liquibase. It would be great if we do not get the `changelog*.xml` files if we generate the project using `flyway`. However, I assume that maven archetype and velocity are too limited to solve this.
So OK if you do not find an easy solution.

templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/DbTestHelper.java
```diff
@@ -1,44 +1,76 @@
...
-  /**
-   * Calls {@link #dropDatabase()} internally, and migrates to the highest available migration (default) or to the
-   * {@code migrationVersion} specified by {@link #setMigrationVersion(String)}.
-   */
-  public void resetDatabase() {
+    try {
+      this.liquibase.dropAll();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
```
hohwille, line 64: same as above. We should take the opportunity to use `TestCleanerPlugin` and avoid any redundant switches between flyway or liquibase in code like this.

templates/server/src/main/resources/archetype-resources/core/src/test/resources/db/changelog/changelog-master.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hohwille, line 1: we have the same file in `src/main/resources` and `src/test/resources`. This IMHO does not make sense.
Like with flyway we should use `db/test` instead of `db/changelog` for tests.
Will it be best practice to use a completely redundant/isolated db config for liquibase for testing vs. main?

templates/server/src/test/resources/projects/basic/archetype.properties
```diff
@@ -5,4 +5,5 @@ groupId=archetype.it
 artifactId=basic
 earProjectName=.
 batch=.
-dbType=h2
\ No newline at end of file
+dbType=h2
+dbMigration=flyway
```
hohwille, line 9: should we also have a test using liquibase to ensure the result builds successfully?

hohwille (changes requested):
@dixyushi thanks for this PR.
You have solved all the tricky problems. However, there are now some cleanups and code-style issues to polish the result before merging. Can you please have a look and rework?

templates/server/src/main/resources/archetype-resources/core/src/main/resources/db/changelog/changelog-master.xml
hohwille: https://stackoverflow.com/questions/20331094/conditionally-include-exclude-files-while-creating-project-using-maven-archetype

templates/pom.xml
dixyushi: I will review it again and restore the file if changes not required

templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
dixyushi: I will fix this steup issue and will make it work.

hohwille (changes requested):
@dixyushi thanks for the improvements. You perfectly catched up with `archetype-post-generate.groovy`. 👍 
However, still the following points are open:
* avoid phantom changes (EOL-style?) for 
  * [ ] `pom.xml`
  * [ ] `templates/pom.xml`
  * [ ] `templates/server/src/main/resources/archetype-resources/__batch__/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java`
* use `TestCleanerPlugin` to abstract from flyway or liquibase in
  * [ ] `templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/DbTestHelper.java`
* Comment from last review not answered for `templates/server/src/main/resources/archetype-resources/core/src/test/resources/db/changelog/changelog-master.xml`
* Exception handling still buggy, see my new review comment.
* Indendation issues, I already updated the idendation issues I found myself as constructive review.

Finally I found a new issue. We have Flyway SQLs for batch. If we want to properly support the batch + liquibase-option we also have to do something here:
https://github.com/devonfw/devon4j/tree/master/templates/server/src/main/resources/archetype-resources/batch/src/main/resources/db/migration/h2
* aa

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/clean/TestCleanerPluginLiquibase.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public void cleanup()  {
+  try {
+    this.liquibase.dropAll();
+  }catch(DatabaseException databaseException) {
+     LOG.error("Database exception occurred!", new DatabaseException());
+  }
+  
+  }
```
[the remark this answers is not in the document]
dixyushi: I have fixed this error handling and passing the same "databaseException" inside LOG.error().

templates/server/src/test/resources/projects/basic/archetype.properties
dixyushi: To do ensure I have added switches here too for liquibase as well as for flyway in all the three archetype properties basic/archetype.properties, batch/archetype.properties and enterprise/archetype.properties.

templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/DbTestHelper.java
dixyushi: Implemented testcleanerPlugin here also to get rid of redundant switches

templates/server/src/main/resources/archetype-resources/core/pom.xml
dixyushi: Using #elseIf was resulting in error i.e "package liquibase doesn't exist" so this is the alternative way which I implemented.

templates/server/src/main/resources/archetype-resources/core/src/test/resources/db/changelog/changelog-master.xml
dixyushi: I have configured this for liquibase separately in Groovy script.

isandesh1986 (approved):

pom.xml
```diff
@@ -1,67 +1,67 @@
...
+      <plugins>
+        <plugin>
+          <groupId>org.springframework.boot</groupId>
+          <artifactId>spring-boot-maven-plugin</artifactId>
+          <version>${spring.boot.version}</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+</project>
```
dixyushi, line 67: I have restored the file but still it is showing in files changed.

templates/pom.xml
```diff
@@ -1,19 +1,19 @@
...
+  <artifactId>devon4j-templates</artifactId>
+  <packaging>pom</packaging>
+  <name>${project.artifactId}</name>
+  <description>Templates (maven archetypes) of devon4j.</description>
+
+  <modules>
+    <module>server</module>
+  </modules>
+
+</project>
```
dixyushi, line 19: I have restored the file  here too but still it is showing in files changed.

modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/clean/TestCleanerPluginNone.java
```diff
@@ -8,7 +8,7 @@
 public class TestCleanerPluginNone implements TestCleanerPlugin {
 
   @Override
-  public void cleanup() {
+  public void cleanup() throws Exception {
```
hohwille, line 11: Do not do this. Please use RuntimeExceptions and avoid checked exceptions wherever possible:
https://github.com/devonfw/devon4j/blob/develop/documentation/guide-exceptions.asciidoc

Also we can not afterwards change APIs in an incompatible manner:
Existing projects would get compiler errors after upgrading devon4j when you do this change.

templates/server/src/main/resources/archetype-resources/batch/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,63 @@
...
+  private TestCleanerPlugin testCleanerPlugin;
+
+  @Override
+  protected void doSetUp() {
+  super.doSetUp();
+#if($dbMigration == 'flyway')
+   testCleanerPlugin = new TestCleanerPluginFlyway();
+#else if($dbMigration == 'liquibase')
+   testCleanerPlugin = new TestCleanerPluginLiquibase();
+#end
```
hohwille, line 35: `testCleanerPlugin` is already annotated with `@Inject`.
So we do we manually have to initialize it here?
IMHO this does not make sense.

templates/server/src/main/resources/archetype-resources/batch/src/test/java/__packageInPathFormat__/general/batch/base/test/SpringBatchIntegrationTest.java
```diff
@@ -1,53 +1,63 @@
...
+try {
+  testCleanerPlugin.cleanup();
+}
+catch(Exception exception) {
+  LOG.error("Exception occurred while performing cleanup", exception);
+}
```
hohwille, line 41: This is what happens due to this checked exception...
Please remove the checked exception and then also this try catch around the cleanup. Catching and further ignoring errors is not a good style if no reasonable compensation of the error can be done.

templates/server/src/main/resources/archetype-resources/core/src/test/java/__packageInPathFormat__/general/common/base/test/DbTestHelper.java
```diff
@@ -1,44 +1,37 @@
...
  * This class provides methods for handling the database during testing where resets (and other operations) may be
  * necessary.
  */
 @Named
-public class DbTestHelper {
+public class DbTestHelper{
 
-  private Flyway flyway;
+  
+  private TestCleanerPlugin testCleanerPlugin;
```
hohwille, line 16: Use `@Inject`. You should get familiar with DI when doing Java coding in devon4j:
https://github.com/devonfw/devon4j/blob/develop/documentation/guide-dependency-injection.asciidoc

hohwille (changes requested):

hohwille:
FYI: https://forum.liquibase.org/t/create-spring-boot-starter/5577

pom.xml
hohwille: Current `pom.xml` has linux EOL style:

```
cat -ben pom.xml
     1	<?xml version="1.0" encoding="UTF-8"?>$
     2	<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"$
     3	  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">$
     4	  <modelVersion>4.0.0</modelVersion>$
....
```

Yours has windows encoding. Maybe something with your git setup or the editor you are using to edit XML files.
```
cat -ben pom.xml.ayushi
     1	<?xml version="1.0" encoding="UTF-8"?>^M$
     2	<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"^M$
     3	  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">^M$
     4	  <modelVersion>4.0.0</modelVersion>^M$
...
```

We once had such changes that added a Byte-Order-Mark to a `pom.xml` that was not easily visible in the diff but broke the release as maven pull-parser can not handle BOM. That is why am a little picky about such things.

Issue #85, opened 2019-05-06
Consider ci-friendly-maven also for archetype


hohwille:
In #74 we introduced [ci-friendly-maven](https://maven.apache.org/maven-ci-friendly.html) for devon4j itself (internal development):
* https://github.com/devonfw/devon4j/blob/master/.mvn/settings.xml
* https://github.com/devonfw/devon4j/blob/master/pom.xml#L30

We should consider to push that out as best-practice to our projects via maven-archetype (`devon4j-template-server`).
https://github.com/devonfw/devon4j/tree/master/templates/server/src/main/resources/archetype-resources

This would also require some nice documentation in our guides. Maybe here:
https://github.com/devonfw/devon4j/blob/master/documentation/guide-structure.asciidoc#project-structure
or
https://github.com/devonfw/devon4j/blob/master/documentation/tutorial-newapp.asciidoc#how-to-run-your-app

hohwille:
Also an old story that should have been fixed for long.
I will have to move to the next release... :(

PR #342, opened 2021-02-08
#85 Consider ci-friendly-maven also for archetype

sujith-mn:


templates/server/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -65,6 +65,14 @@
         <include>.gitignore</include>
       </includes>
     </fileSet>
+	
+	<fileSet packaged="false" filtered="true" encoding="UTF-8">
+      <directory>.mvn</directory>
+      <includes>
+        <include>**/*.*</include>
+      </includes>
+    </fileSet>
```
hohwille, line 74: Can you fix the indendation (here, and in all POMs you changed)... thanks.

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -19,6 +19,7 @@
     <jackson.version>$[jackson.version]</jackson.version> <!-- Overriding Jackson for fixing vulnerabilities -->
     <guava.version>$[guava.version]</guava.version>
     <devonfw.test.excluded.groups>system</devonfw.test.excluded.groups>
+	 <app.version>${revision}${changelist}</app.version>
```
hohwille, line 22: Excellent, this is exactly what we wanted to archive.
Please fix indendation and we can merge.

templates/server/src/main/resources/archetype-resources/.mvn/maven.config
```diff
@@ -0,0 +1 @@
+-Drevision=0.0.1 -Dchangelist=-SNAPSHOT
```
hohwille, line 1: An archetype can take a version as argument. That is available via `${version}` variable.
It will now be ignored and the initial version is hardcoded to this `0.0.1-SNAPSHOT`.
As `${version}` can also be `1.0.0-SNAPSHOT` we can not easily do
```
-Drevision=${version} -Dchangelist=-SNAPSHOT
```

As this would then result in `1.0.0-SNAPSHOT-SNAPSHOT`.
However, as a nice to have, we could check if via velocity macros it is possible to strip out a potential `-SNAPSHOT` from `${version}`. If that is possible, we should use `${version}` to make it perfect.
However, for me also fine to approve and merge like this.

hohwille (changes requested):
@sujith-mn Thanks for implementing this story. You properly solved this and I wonder why we havent been able to solve this earlier or what have been the problems in the first place. Anyhow it is great that you made it and we finally get this done 👍 

Can you please fix the indendation (see our DoD) so I can approve and merge?
I also added some nice to have imporvement that you can have a look at. But that can be done after merging this PR or never if not easily doable.

templates/server/src/main/resources/archetype-resources/.mvn/maven.config
sujith-mn: The formatting is done..

version passing as an argument would be really nice..

hohwille: Thanks for formatting. To prevent blocking this PR, I will merge it now.
Can you please create a new issue for the nice-to-have improvement to put the version here and research velocity macro syntax to remove potential `-SNAPSHOT` suffix?

templates/server/src/main/resources/archetype-resources/pom.xml
```diff
@@ -19,17 +19,18 @@
...
-#if ($earProjectName != '.')
+    #if ($earProjectName != '.')
     <module>${earProjectName}</module>
-#end
-#if ($batch == 'batch')
+    #end
+    #if ($batch == 'batch')
     <module>batch</module>
-#end
+    #end
```
hohwille, line 33: Seems Eclipse formatter also indents velocity macros.
We need to check the generated output as this might break the indentation there.
I hoped eclipse would leave things like this as is...
If this is a problem in the resulting pom we can fix in the new story for improvement.

hohwille (approved):
Thanks. Ready to merge now.

templates/server/src/main/resources/archetype-resources/.mvn/maven.config
sujith-mn: Issue #346 created.

is it okay to directly place the version to app.version properties ... like below

#if($version)
	<app.version>${version}</app.version>
#else
	<app.version>${revision}${changelist}</app.version>
#end

PR #300, opened 2020-09-24
#294 : Sonar integration with devon4j

suvmanda:
Fix for issue #294 sonarcloud integration.

Issue #294, opened 2020-09-01
integrate sonarcloud analysis into devon4j CI pipeline


hohwille:
As a product-owner of devon4j, I want to have sonarcloud analysis integrated into our CI so that general code quality can be assured easily.

For `sonar-devon4j-plugin` this has already been implemented:
https://github.com/devonfw/sonar-devon4j-plugin/blob/abc5bc245fe1ba1e5be5876cba97668d65a46a40/.travis.yml#L13

The same shall also be done for `devon4j`. This might imply some general code-cleanup to satisfy with SonarQube default quality profile.
Please notice issue #347 and PR #350

hohwille:
LOL:
I followed this guide:
https://github.com/SonarSource/sonarcloud-github-action
But now I get this error:
```
Maven project detected. You should run the goal 'org.sonarsource.scanner.maven:sonar' during build rather than using this GitHub Action.
```

hohwille:
Next error:
```
Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.7.0.1746:sonar (default-cli) on project devon4j: Could not find a default branch to fall back on.
```

hohwille:
Fixed some more errors. Documentation is wrong/incomplete.
Next error now is:
```
Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.7.0.1746:sonar (default-cli) on project devon4j: Unable to execute SonarQube: Fail to get bootstrap index from server: Failed to connect to localhost/0:0:0:0:0:0:0:1:9000: Connection refused (Connection refused) -> [Help 1]
```

It seems that the sonar-cloud URL is not used and has to be explicitly configured. I am slightly confused by the documentation of sonarcloud that does not address all these problems...

hohwille:
Wow, now after fixing all the properties I finally get this error:
```
Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.7.0.1746:sonar (default-cli) on project devon4j: null: MojoExecutionException: NullPointerException
```
Nice.

hohwille:
Hurray, here comes the stacktrace:
```
Caused by: org.apache.maven.plugin.MojoExecutionException
     at org.sonarsource.scanner.maven.bootstrap.ScannerBootstrapper.execute (ScannerBootstrapper.java:67)
     at org.sonarsource.scanner.maven.SonarQubeMojo.execute (SonarQubeMojo.java:104)
[...]

PR #343, opened 2021-02-09
#347: github actions for devon4j

ssarmokadam:
Provides github action config for issue #347


hohwille (approved):
thanks @ssarmokadam for this PR.
Looks good to me. We will immediately try it and I therefore merge it.
If all works as expected, we can disable travis build.

One thing to notice is that PR #300 / PR #299 is now obsolete, needs to be closed and redone here for github action.

hohwille (commented):
One further question: where are the secrets for deployment? IMHO this cannot work yet?!?

hohwille (changes requested):
I reviewed deeper and tried to understand more of github actions and now see two problems:
* the secrets for deployment are missing so deployment will fail
* github actions does not provide repo/project secrets for PR builds. While this is a security feature preventing users from creating a PR that echos secrets in the build log, it prevents us from implementing #294 and therefore I am questioning if the migration from travis to github actions makes sense at all...

hohwille:
OK, as I commented in #347 there seems to be a way to get all working with github action.
I try to be agile so I merge this PR already and expect another PR for #347 to improve but we can already get started and activate github actions in parallel to travis...

hohwille:
I added secrets for SONATYPE_USER and SONATYPE_PASSWORD.
The github action build failed with this error:
```
fatal: unable to access 'https://github.com/${ORG}/${REPO_DEST}.git/': The requested URL returned error: 400
Error: Process completed with exit code 128.
```

ssarmokadam:
@hohwille I will check and create new PR with changes required

hohwille:
You can also just update this PR.

Issue #347, opened 2021-02-16
migrate from travis to github actions


hohwille:
For several reasons we have finally decided to migrate from travis-ci to github actions.

hohwille:
I just reviewed PR #343 and was confused about the secrets. Now I learned that we can not have secrets for PR builds. This will prevent us from implementing PR #294 so I am not convinced that we should go away from travis if we can not solve this with github actions...

hohwille:
OK, it seems possible after all:
* https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
* https://github.com/SonarSource/sonarcloud-github-action

So let us try this out...

hohwille:
@ssarmokadam can you please asap get in touch with @sujith-mn 
He has created the github actions for devonfw-ide:
https://github.com/devonfw/ide/pull/529/files

We do not seem to be aligned here and follow different approaches and conventions.

maybeec:
@ssarmokadam , @sujith-mn , please check how it is done in solicitor as we discussed it might be the most complete and working setup. Please make sure the files apply the same standards and names. Even I just made quite some reviews and sometimes there is the email coming from the secrets, sometimes not. It should be same for all. Please make sure, we have the same view on it and have a continuous chat on that do not do the work twice ;)

hohwille:
Was autoclosed by PR. Will reopen as not yet done.
Only after all working smoth travis disabled and  `.travis.yml` being deleted, we will close this issue.

hohwille:
@ssarmokadam have you talked with @sujith-mn about this?
I have completed github actions in devonfw-ide. You can get the changes from there to complete this in devon4j.
https://github.com/devonfw/ide/pull/533

hohwille:
Reopened due to the review comment in #360 that suggests a final improvement.

hohwille:
We still have the old github action in place: `deploy.yml` was not renamed to `build.yml` but instead has been copied.
So we need to delete `deploy.yml`...

hohwille:
I just fixed the badge that has been forgotten to change and still was pointing to travis.
However, still this does not seem to work as it also runs in PRs:
https://github.com/devonfw/devon4j/actions/runs/675514421/workflow#L26

ssarmokadam:
@hohwille It looks like access issue ... SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
[...]

PR #350, opened 2021-02-18
#347: github action for devon4j

ssarmokadam:
@hohwille  I have discussed this with @sujith-mn and we agreed to this format. Please let me know any further changes are required specially for documentation step as we can not test it in fork.

Provides github action config for issue #347


hohwille:
See https://github.com/devonfw/ide/pull/529/files and https://github.com/devonfw/solicitor/tree/master/.github/workflows
Should be `sync-docs.yml` for the synchronization of the documentation to the wiki.

hohwille (approved):
@ssarmokadam thanks for this PR. Looks good to me now 👍 
Lets merge and observere the results when running the actions.

PR #351, opened 2021-02-18
improved error handling of service client

hohwille:
The service client shoud report errors comming from spring boot rather than our own `RestServiceExceptionFacade` properly and report HTTP status code and error properly via exception.

### DIFF
--- a/modules/service/src/main/java/com/devonfw/module/service/common/base/client/AbstractServiceClientErrorFactory.java
+++ b/modules/service/src/main/java/com/devonfw/module/service/common/base/client/AbstractServiceClientErrorFactory.java
@@ -71,10 +71,24 @@ public abstract class AbstractServiceClientErrorFactory implements ServiceClient
   private RuntimeException createException(Map<String, Object> map, String service) {
 
     String code = (String) map.get(ServiceConstants.KEY_CODE);
+    if (isEmpty(code)) {
+      code = map.get(ServiceConstants.KEY_STATUS).toString();
+    }
     String message = (String) map.get(ServiceConstants.KEY_MESSAGE);
+    if (isEmpty(message)) {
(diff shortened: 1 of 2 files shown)


modules/service/src/main/java/com/devonfw/module/service/common/base/client/AbstractServiceClientErrorFactory.java
```diff
@@ -71,10 +71,24 @@ public RuntimeException unmarshall(String data, String format, int statusCode, S
   private RuntimeException createException(Map<String, Object> map, String service) {
 
     String code = (String) map.get(ServiceConstants.KEY_CODE);
+    if (isEmpty(code)) {
+      code = map.get(ServiceConstants.KEY_STATUS).toString();
```
maybeec, line 75: this could yield to a NPE!?

maybeec (changes requested):

modules/service/src/main/java/com/devonfw/module/service/common/base/client/AbstractServiceClientErrorFactory.java
hohwille: ??? Your comment does not make sense to me. That is exactly the purpose of the `isEmtpy` method to prevent NPE. Did you read the code?

maybeec: `map.get(ServiceConstants.KEY_STATUS)` could be null right?

hohwille: Ouch. Sure. Sorry, I was stupid and missed this. Will fix this tomorrow. Thanks...

maybeec (approved):

Issue #346, opened 2021-02-16
Version variable ignored by maven archetype


sujith-mn:
See issue #85 / PR #342

An archetype can take a version as argument. That is available via ${version} variable.
It will now be ignored and the initial version is hardcoded to this 0.0.1-SNAPSHOT.
As ${version} can also be 1.0.0-SNAPSHOT we can not easily do

-Drevision=${version} -Dchangelist=-SNAPSHOT
As this would then result in 1.0.0-SNAPSHOT-SNAPSHOT.
However, as a nice to have, we could check if via velocity macros it is possible to strip out a potential -SNAPSHOT from ${version}. If that is possible, we should use ${version} to make it perfect.

hohwille:
https://github.com/devonfw/devon4j/blob/89ce5fdbce869748fe596a3e0b28e1ae143ac5b7/templates/server/.mvn/maven.config#L1

hohwille:
Seems easy:
https://help.mypurecloud.com/articles/velocity-macros-data-actions/#JavaStringMethods

sujith-mn:
devon4j template gives the below error while building

INFO] [INFO] enterprise-server 0.0.1-SNAPSHOT ................... SKIPPED
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] BUILD FAILURE
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] Total time:  01:17 min
[INFO] [INFO] Finished at: 2021-02-18T11:31:38+05:30
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [ERROR] Failed to execute goal on project enterprise-ear: Could not resolve dependencies for project archetype.it:enterprise-ear:ear:1.0.0-SNAPSHOT: Failed to collect dependencies at archetype.it:enterprise-server:war:1.0.0-SNAPSHOT: Failed to read artifact descriptor for archetype.it:enterprise-server:war:1.0.0-SNAPSHOT: Failure to find archetype.it:enterprise:pom:${app.version} in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
[...]

PR #353, opened 2021-02-25
#346 Version variable ignored by maven archetype

sujith-mn:
Addresses #346


hohwille (approved):
@sujith-mn thanks for this PR.
I will still have to rework the `maven.config` but this is a good improvement. 👍 
We could also follow @maybeec suggestion and simply remove and ignore `changelist` to make it simply. I will change devonfw-ide `release` commandlet to also support this.

PR #359, opened 2021-03-05
removed the special character from the pom.xml

sujith-mn:


templates/server/src/main/resources/archetype-resources/core/pom.xml
```diff
@@ -122,50 +122,50 @@
 
     <!-- Database -->
     <dependency>
-#if ($dbType == 'h2')
+      #if ($dbType == 'h2')
```
hohwille, line 125: it is great that you do auto-formatting now.
Unfortunately in the archetype with velocity macros this is not so great as the extra whitespaces in front of macros will be written to the output. Therefore the generated POM will be messed with whitespaces. Can you have a look at the resulting generated POM?

hohwille:
@sujith-mn So we can close this PR without merging?

sujith-mn:
> @sujith-mn So we can close this PR without merging?

Merging is not required, did not see any compile issue now.

hohwille:
So to finally get things done if the review is not progressed, I redid the changes in a new PR #379.

PR #360, opened 2021-03-11
updating github actions for devon4j as ide

ssarmokadam:


.github/workflows/sync-docs.yml
```diff
@@ -1,4 +1,5 @@
-name: github actions for devon4j wiki
+name: Sync Docs
```
hohwille, line 1: I still do not get why the GH action config in `devonfw-ide` that does the exact same thing looks very different...

hohwille (approved):
@ssarmokadam thanks for the PR. GH actions look functionally correct.
Therefore I approve and we can merge.
However, the issue is IMHO still not done as our intention is to have the actions aligned accross the repositories and `sync-docs.yml` has many differences here compared to IDE that do not make any sense to me.

.github/workflows/sync-docs.yml
ssarmokadam: @hohwille in ide below 3 lines are added instead of defining ORG, REPO_SOURCE and REPO_DEST in env variable. I will do the same for devon4j to align with ide actions. Apart from this other things are same 

   echo "ORG=$(echo '${{ github.repository }}' | awk -F '/' '{print $1}')" >> $GITHUB_ENV
   echo "REPO_SOURCE=$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')" >> $GITHUB_ENV
   echo "REPO_DEST=$(echo '${{ github.repository }}' | awk -F '/' '{print $2}').wiki" >> $GITHUB_ENV

PR #363, opened 2021-03-15
#294: sonarcloud

hohwille:
Next try for #294 (converted to maven build).


hohwille:
Still the same NPE.

hohwille:
To test if this is a PR limitation of accessing the secrets (`SONAR_TOKEN`), I will merge this and then we can see if it still fails with the same error (what I assume). If that is the case, I will ask for help in the sonar forum.

PR #364, opened 2021-03-16
#347: #294: improved workflows and sonaranalysis

hohwille:
Improvements for #347 and #294:
* merge sonar into build workflow
* use `ci` profile for extra steps
* split build into 3 steps
* allow secrets and sonarcloud for PRs.


hohwille:
`build` seems to hang now. Maybe because of https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target that first has to be comitted to default branch...

PR #365, opened 2021-03-19
Bugfixing initialized flag in BaseTest

PhilippHedwig:
As the initialized flag was declared static, it was used as an indicator for first initialization of any test (using BaseTest), not only for the test class. To make it work as it was intended, 'static' needs to be removed. I renamed the property from upper to lower case to reflect that it's not static.

Also, the method isInitialSetup() inverts the meaning of the flag, but simply uses it directly. The class is in it's initial state when it wasn't initialized before. To fix this, I introduced inverting the flag in the method.

I hope I didn't break anything with editing this directly in GitHub without IDE support. This especially means I wasn't able to test the change. As the flag is declared protected, it might be used directly in other classes.

### DIFF
--- a/modules/test/src/main/java/com/devonfw/module/test/common/base/BaseTest.java
+++ b/modules/test/src/main/java/com/devonfw/module/test/common/base/BaseTest.java
@@ -43,7 +43,7 @@ public abstract class BaseTest extends Assertions {
    * Indicates if the test class is to be set up for the first time. {@code true} indicates that the class has already
    * been set up (e.g., database setup) for the execution of an preceding test method.
    */
-  protected static boolean INITIALIZED = false;
+  protected boolean initialized = false;
 
   /**
    * Suggests to use {@link #doSetUp()} method before each tests.
@@ -53,8 +53,8 @@ public abstract class BaseTest extends Assertions {
 
     // Simply sets INITIALIZED to true when setUp is called for the first time.
     doSetUp();
-    if (!INITIALIZED) {
-      INITIALIZED = true;
+    if (!initialized) {
+      initialized = true;
     }
   }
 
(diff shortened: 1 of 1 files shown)


modules/test/src/main/java/com/devonfw/module/test/common/base/BaseTest.java
```diff
@@ -43,7 +43,7 @@
    * Indicates if the test class is to be set up for the first time. {@code true} indicates that the class has already
    * been set up (e.g., database setup) for the execution of an preceding test method.
    */
-  protected static boolean INITIALIZED = false;
+  protected boolean initialized = false;
```
hohwille, line 46: You can not solve this non-static:
JUnit creates a new instance of the class for each test run (method).
Otherwise this problem would have been very easy to solve.

modules/test/src/main/java/com/devonfw/module/test/common/base/BaseTest.java
```diff
@@ -73,7 +73,7 @@ public final void tearDown() {
    */
   protected boolean isInitialSetup() {
 
-    return INITIALIZED;
+    return !initialized;
```
hohwille, line 76: Indeed this was inverse and has always been wrong. Seems I missed this in my review and it is a shame that nobody found all this earlier...

hohwille (changes requested):
@PhilippHedwig Thanks for your PR.
However, this is not fixing the problem as stated in my review-comments.
But thank you very much for this finding.
We should first create an issue for this bug and after that come up with a PR as fix.
The design of JUnit4 makes it not so easy to solve this but it is a shame that this feature has never been tested properly in devon4j.
I see two options to properly fix this but I will rather comment the bug issue that I am going to create.

modules/test/src/main/java/com/devonfw/module/test/common/base/BaseTest.java
PhilippHedwig: JUnit creating a new instance for every test was actually new to me and you're right, this doesn't fix the problem.

hohwille:
Replaced by PR #367

PR #368, opened 2021-04-08
#346 removed changelist from the version.

sujith-mn:


templates/server/.mvn/maven.config
```diff
@@ -1,6 +1,5 @@
 #if($version)
-	#set ($versionArray = $version.split("-"))
-	-Drevision=$versionArray.get(0) -Dchangelist=-$versionArray.get(1)
+	-Drevision=$version
```
hohwille, line 2: fine but isnt `$version` always set? Then why the entire if and else?

hohwille: Why do we have this twice (`templates/server/.mvn/maven.config` and `templates/server/src/main/resources/archetype-resources/.mvn/maven.config`)?
IMHO this file is never used and can be deleted.

templates/server/src/main/resources/archetype-resources/.mvn/maven.config
```diff
@@ -1 +1,5 @@
--Drevision=0.0.1 -Dchangelist=-SNAPSHOT
+#if($version)
```
hohwille, line 1: see comment above. IMHO the if is pointless.

hohwille (approved):
@sujith-mn thanks for this PR. Looking good and simplifies the entire thing. 👍 
We could then also do the same in devon4j (and IDE) itself - but with separate PRs.
Can you have a look at my review comments and imporve to make it clean?

hohwille:
> We could then also do the same in devon4j (and IDE) itself - but with separate PRs.

Created issue #369 in devon4j

hohwille (approved):
Also #369 is now properly addressed. Thanks @sujith-mn - great work 👍

PR #387, opened 2021-04-29
rebuild and updated diagram with drawio

hohwille:
access control diagram redone with drawio


hohwille:
I do not see this critical and IMHO it is an obvious improvement so I am no longer waiting for a review...

PR #392, opened 2021-05-19
#336 archetype optional batch

### LINKED ISSUE #336: archetype contains batch artefacts even when no batch was generated
### Expected behavior

As a user of devon4j, I want to create a new app with minimal content so that I am not spammed with unneccessary pointless classes that I have to delete.

### Actual behavior

Even though I did not choose to generate my app with batch I got the following files (starting from the root of the create app and *NOT* in `batch` subfolder/module):
* `./src/main/resources/db/migration/h2/V0005__Add_batch_tables.sql`
* `./src/main/java/org/example/app/SpringBootBatchApp.java`
* `./src/main/java/org/example/app/general/batch/impl/config/BeansBatchConfig.java`
* `./src/test/java/org/example/app/general/batch/base/test/SpringBatchIntegrationTest.java`
* `./src/test/java/org/example/app/general/common/base/test/TestUtil.java`

This looks like something is messed in the archetype to me.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. `devon java create org.example.app`
2. `cd app`
3. `find src`

### Affected version:

Latest `devon4j` version `2020.12.001`.

tobka777:
fixes #336


hohwille:
I reviewed PR #394 and think it is a better solution. Sorry, for the extra work, but I only found this solution very recently. I will therefore close this PR.

PR #394, opened 2021-05-27
#336 archetype optional batch with post-generation script

### LINKED ISSUE #336: archetype contains batch artefacts even when no batch was generated

tobka777:
fixes #336


templates/server/src/main/resources/META-INF/archetype-post-generate.groovy
```diff
@@ -0,0 +1,30 @@
...
+import java.nio.file.LinkOption
+
+// the path where the project got generated
+Path projectPath = Paths.get(request.outputDirectory, request.artifactId)
+
+// the properties available to the archetype
+Properties properties = request.properties
+String batch = properties.get("batch")
+
+if (batch != "batch") {
```
hohwille, line 13: OK, indeed in goovy this is using equals, so fine :)

templates/server/src/main/resources/META-INF/archetype-post-generate.groovy
```diff
@@ -0,0 +1,30 @@
...
+}
+
+void deleteDirectoryRecursion(Path path) {
+  if (Files.isDirectory(path, LinkOption.NOFOLLOW_LINKS)) {
+    try {
+      for (Path entry : Files.newDirectoryStream(path)) {
+        deleteDirectoryRecursion(entry);
+      }
+    } catch(IOException e) {
+      e.printStackTrace()
```
hohwille, line 26: surely an exotic situation but we should avoid using `printStackTrace()`.
IMHO simply wrap the IOException into a new RuntimeException and throw that one to abort the process in case of unexepcted I/O error.

templates/server/src/main/resources/META-INF/archetype-post-generate.groovy
```diff
@@ -0,0 +1,30 @@
...
+  // delete folder batch recursively
+  Path rootPath = projectPath.resolve("batch")
+  deleteDirectoryRecursion(rootPath)
+}
+
+void deleteDirectoryRecursion(Path path) {
+  if (Files.isDirectory(path, LinkOption.NOFOLLOW_LINKS)) {
+    try {
+      for (Path entry : Files.newDirectoryStream(path)) {
+        deleteDirectoryRecursion(entry);
```
hohwille, line 23: as you make use of omitting semicolons as it seems best practice in groovy, it would be even nicer, if you do that consistently (including line 29).

hohwille (approved):
@tobka777 thanks for this PR. This solution is nice and causes minimal impact to fix #336. 👍 
I added some minimum review comments. As they are really nice to have, I already approve this PR. If you happen to see this quickly and can rework this would be great. Otherwise we can also merge as is and do these minimal improvements later. After all this PR solves #336 properly in merging it makes devon4j better :)

Issue #216, opened 2020-02-18
ability to generate OpenApi files automatically from JAX-RS services


StevenGarcia-CG:
As a developer I want to be able to have an automated swagger.json file generation, at present it's not possible with the JAX-RS implementation that devon4J offers

StevenGarcia-CG:
There may be a solution to this, if we follow this cookbook done by @AbhayChandel :  https://github.com/devonfw-forge/devon-guide/wiki/Cookbook-Swagger

hohwille:
Some time ago we had demand for generation of (swagger like) documentation for JAX-RS REST service interfaces. I spent two weekends and wrote this plugin:
https://github.com/mojohaus/servicedocgen-maven-plugin/
It should be relatively easy to extend this with the ability to already write the metamodel into a swagger.json file.
Of course the plugin should also be modified so it can not only be used for reporting (maven site generation) but also as regular plugin to generate standalone.
My problem is that I am very busy and continuously overloaded and do not expect to have time for this within the next couple of months. PRs are of course welcome, but in case I would have to do it, we need some special prioritization in devonfw...

hohwille:
https://www.mojohaus.org/servicedocgen-maven-plugin/faq.html

> Can you also generate swagger.json?
>    No. Not yet. Please get involved and create a pull-request if you want to have this.

hohwille:
https://github.com/mojohaus/servicedocgen-maven-plugin/issues/25

hohwille:
moving to the next release as PR is still stuck.

hohwille:
servicedocgen-maven-plugin has been updated accordingly.
However, shipping a release is failing:
https://groups.google.com/g/mojohaus-dev/c/CpjeLohpeAY

hohwille:
Release performed and just staged. Should shortly appear here:
https://repo1.maven.org/maven2/org/codehaus/mojo/servicedocgen-maven-plugin/

hohwille:
servicedocgen 1.0.0 is on its way:
https://groups.google.com/g/mojohaus-dev/c/Kchz1vUcqJA

I also did some advanced testing and found+fixed an important bug.
After update in devon4j to 1.0.0 we can close this issue and IMHO have a good solution.

hohwille:
Release 1.0.0 is now available with bugfixes and swagger UI support:
https://repo1.maven.org/maven2/org/codehaus/mojo/servicedocgen-maven-plugin/1.0.0/

PR #395, opened 2021-05-31
#216: update servicedoc and add to regular build

hohwille:
implements #216:
* updated parent to update servicedocgen to beta6
* generate servicedoc not only in site but also in regular build
* added build for all modules (not only API) for more flexibility on purpose (if projects want to avoid REST interface in `api`).


maybeec (approved):

hohwille:
Seems as if this PR was overtaken by PR #405, I merged it with master and nothing left except phase.
However as this phase seems to be more appropirate I will also merge this PR.

Issue #385, opened 2021-04-29
Access-control should honor roles by default


hohwille:
devon4j ships with an application template that already comes with a login mechanism and a ready to use function based access control for authorization.
However, there are two places that do not seem to fit together:
1. https://github.com/devonfw/devon4j/blob/92a6ec127b90542550c2ecd6f2615ee28200bb2e/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/service/impl/config/BaseWebSecurityConfig.java#L115
2. https://github.com/devonfw/devon4j/blob/92a6ec127b90542550c2ecd6f2615ee28200bb2e/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java#L86

So with 1. we also assign a role to a user.
However, with 2. we ignore this role assignment and simply take the login username as role what is obviously nuts (we already have a todo comment on top of this).
Aint there a servlet JEE api to get the roles of the user that we could use here? I assume the roles assigned via spring-security fluent API can also be retrieved somehow.

tobka777:
I have looked at it. Wouldn't you just have to replace this line 
https://github.com/devonfw/devon4j/blob/92a6ec127b90542550c2ecd6f2615ee28200bb2e/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java#L48
with
`User userData = new User(user.getUsername(), user.getPassword(), user.getAuthorities());`

`user.getAuthorities()` returns for user admin [ROLE_Admin]. 
And we don't need the methods getAuthorities and getRoles.

Maybe I am also wrong.

hohwille:
Sounds good 👍 
Sometimes the solutions are easy ;)

hohwille:
One concern:
The `ROLE_` prefix is undesired. We already introduced some solution for spring-security to avoid adding this by default.
So we should also strip it off here if present.

hohwille:
https://github.com/devonfw/devon4j/blob/682c72e6440c9172806ae6393c553b2dc1853459/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/logic/impl/config/DefaultRolesPrefixPostProcessor.java#L16

[...]

PR #398, opened 2021-06-15
Issue#385: ROLE prefix removed

ssarmokadam:


templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
```diff
@@ -60,33 +55,31 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
...
+        .collect(Collectors.toUnmodifiableSet());
+
+    undefinedIds.forEach(id -> LOG.warn("Undefined access control {}.", id));
+
+    return accessControlSet.stream().map(accessControl -> new AccessControlGrantedAuthority(accessControl))
+        .collect(Collectors.toUnmodifiableSet());
   }
 
-  private Collection<String> getRoles(String username) {
+  static final String ROLE_PREFIX = "ROLE_";
```
hohwille, line 73: We could find a better spot making this constant public but already fine if it does not cause redundancies.

hohwille (approved):
@ssarmokadam Excellent work 👍 
Also good catch that you removed the PREFIX from the Group what was actually wrong before. Seems as you have properly tested and/or thought about everything.

As discussed in #385 we should have another PR for improving the removal of the ROLE_ prefix from spring-security.

Issue #284, opened 2020-08-21
Support for synchronous rest client using HTTP client


hohwille:
For #279 we figured out that none of the existing JAX-RS implementations (CXF, jersey, etc.) have proper async support that allows to use the JAX-RS annotated service interface in a type-safe way.
Therefore, we created PR #280 that implements this from scratch based on Javas new, efficient and reactive HTTP client.
As we had to solve all the things and complexity of JAX-RS (at least for typical use-cases) and we have to maintain this in `devon4j` we could easily offer a synchronous implementation this way as well.
This would then be an alternative to `devon4j-cxf-client-rest` and would allow projects to reduce dependencies (CXF and all its transitive deps).

hohwille:
As we are overdue with the release and no PR is approaching, I will have to move this story to the next release.

PR #401, opened 2021-06-16
Feature/284 support for synchronous rest client using http client

suvmanda:
Implements issue #284


modules/http-client-rest/src/main/java/com/devonfw/module/httpclient/common/impl/rest/HttpRestClientSyncAutoConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import com.devonfw.module.service.common.api.client.sync.SyncHttpServiceClientFactory;
+import com.devonfw.module.service.common.impl.client.ServiceClientAutoConfiguration;
+
+/**
+ * {@link Configuration} for REST (JAX-RS) clients using Java HTTP client.
+ *
+ * @since 2020.08.001
```
ssarmokadam, line 13: check from which version feature is available and modify javadoc accordingly

modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractSyncServiceHttpClient.java
```diff
@@ -0,0 +1,135 @@
...
+  @SuppressWarnings("null")
+  @Override
+  protected <R> void doCall(ServiceClientInvocation<S> invocation, Consumer<R> resultHandler) {
+
+    long startTime = System.nanoTime();
+    HttpRequest request = createRequest(invocation);
+    HttpResponse<String> future = null;
+    try {
+      future = this.client.getHttpClient().send(request, BodyHandlers.ofString());
+    } catch (IOException e) {
```
ssarmokadam, line 55: Check for exception handling part in asynclient . do not add e.printStackTrace()

modules/service/src/main/java/com/devonfw/module/service/common/api/client/sync/SyncServiceClientFactory.java
```diff
@@ -25,4 +25,14 @@
    */
   <S> S create(ServiceContext<S> context);
 
+  /**
```
ssarmokadam, line 28: Remove commented code

ssarmokadam (changes requested):
Check for javadocs on method and modify it for synchronous, check for versions in doc. Also check for error handling part

modules/service/src/main/java/com/devonfw/module/service/common/api/client/ServiceClientFactory.java
```diff
@@ -79,4 +79,23 @@
...
+
+  <S> SyncServiceClient<S> createSync(Class<S> serviceInterface, Map<String, String> config);
+
+  /**
+   * @param <S> type of the {@link com.devonfw.module.service.common.api.client.context.ServiceContext#getApi() service
+   *        API}.
+   * @param serviceInterface the {@link Class} reflecting the interface that defines the API of your {@link Service}.
+   * @return an {@link SyncServiceClient} allowing to call service operations synchronously.
+   */
+  <S> SyncServiceClient<S> createSync(Class<S> serviceInterface);
```
hohwille, line 99: We already have this API. For historic reasons the methods are called `create`. When async support was added the `createAsync` methods had been added.
I do not see why we have to add `createSync` additionally that would IMHO do the same as the `create` methods. Do I miss something?

modules/service/src/main/java/com/devonfw/module/service/common/api/client/SyncServiceClient.java
```diff
@@ -0,0 +1,77 @@
...
+ * default like {@code 0} or {@code false}). This approach allows the {@link #call(Object, Consumer) call} method to be
+ * type-safe ensuring that the {@link Consumer} callback fits to the result type of the invoked operation.<br>
+ * Therefore ensure that you always call exactly one service operation on the {@link #get() service client} immediately
+ * followed by a {@link #call(Object, Consumer) call} method invocation.
+ *
+ * @param <S> type of the {@link #get() service client}.
+ * @since 2021.08.003
+ * @see ServiceClientFactory#createSync(Class)
+ */
+public interface SyncServiceClient<S> {
```
hohwille, line 31: OK. It seems there is a missunderstanding here.
The synchronous access using HTTP should behave exactly like the API we already have existing.
Only the a new implementation option should be added that can be used instead without `cxf-client` dependencies...

IMHO for synchronous invocation this new API seems overcomplicated. I do not have to provide `Consumer` as callback when I can call the code directly.

hohwille (changes requested):
@suvmanda Thanks for this PR. You have done a great job here. The code looks very well, has nice JavaDoc with `since` tags, code-style is fine. 👍 

However, you extended the API what seems to be a missunderstanding. Can you please check this and if you agree rework your PR?

modules/service/src/main/java/com/devonfw/module/service/common/api/client/SyncServiceClient.java
hohwille: See existing documentation:
https://github.com/devonfw/devon4j/blob/master/documentation/guide-service-client.asciidoc#simple-usage

```
 private void callSynchronous(MyArgs myArgs) {
    MyService myService = this.serviceClientFactory.create(MyService.class);
    // call of service over the wire, synchronously blocking until result is received or error occurred
    MyResult myResult = myService.myMethod(myArgs);
    handleResult(myResult);
  }
```

So when replacing dependency `devon4j-starter-cxf-client-rest` with `devon4j-starter-http-client-rest-sync` the same code pattern should work except that under the hood no more CXF but the HttpClient is used.

modules/http-client-rest/src/main/java/com/devonfw/module/httpclient/common/impl/rest/SyncServiceClientStubImpl.java
```diff
@@ -0,0 +1,306 @@
...
+  private Consumer<Throwable> errorHandler;
+
+  /** The most recent invocation. */
+  private ServiceClientInvocation<S> invocation;
+
+  private ServiceClientStub<S> stub;
+
+  /** The owning {@link SyncServiceClientFactoryHttp factory} which created this client. */
+  // protected final F factory = null;
+
```
hohwille, line 63: remove outcommented code.

modules/http-client-rest/src/main/java/com/devonfw/module/httpclient/common/impl/rest/SyncServiceClientStubImpl.java
```diff
@@ -0,0 +1,306 @@
...
+    String contentType = response.headers().firstValue("Content-Type").orElse("application/json");
+    String data = "";
+    Object body = response.body();
+    if (body instanceof String) {
+      data = (String) body;
+    } else {
+      handleUnsupportedBody(body);
+    }
+    // return this.factory.getErrorUnmarshaller().unmarshall(data, contentType, statusCode, service);
+    return null;
```
hohwille, line 139: Please complete this with proper code.

modules/http-client-rest/src/main/java/com/devonfw/module/httpclient/common/impl/rest/SyncServiceClientStubImpl.java
```diff
@@ -0,0 +1,306 @@
...
+
+  /** {@link ServiceHttpClient} to use. */
+  protected final ServiceHttpClient client;
+
+  private final ObjectMapper objectMapper;
+
+  private final RestServiceMetadata<S> serviceMetadata;
+
+  /** The {@link #setErrorHandler(Consumer)} */
+  private Consumer<Throwable> errorHandler;
```
hohwille, line 54: should also be removed as this only  makes sense for async usage. There is no API to set this in sync usage and the sync API is designed to throw errors immediately when received.

modules/http-client/src/main/java/com/devonfw/module/httpclient/common/impl/AbstractSyncServiceHttpClient.java
```diff
@@ -0,0 +1,76 @@
...
+import com.devonfw.module.service.common.base.client.AbstractSyncServiceClient;
+
+/**
+ * Abstract base implementation of {@link SyncServiceClient} using Java HTTP client.
+ *
+ * @param <S> type of the {@link #get() service client}.
+ * @param <F> type of the owning {@link SyncServiceClientFactoryHttp factory}.
+ * @since 2021.08.003
+ */
+
```
hohwille, line 18: can you remove all these pointless empty lines between JavaDoc and Signature, please?

hohwille (changes requested):
@suvmanda thanks for your updates. We are getting close. Please have a look at my review comments.
There has been some C&P from the existing code. I can resolve that myself after merging.

hohwille:
@suvmanda thanks again for your work and this PR. I forked your work and created PR #410 that replaces this PR.
Therefore I am closing this PR.

PR #402, opened 2021-06-16
#400: no ROLE_ prefix using authorities

### LINKED ISSUE #400: Remove the ROLE_ prefix from Spring Security using JavaConfig
As a Developer, I want to remove ROLE_ using java config.

Refer: https://stackoverflow.com/questions/38134121/how-do-i-remove-the-role-prefix-from-spring-security-with-javaconfig

Related Issue: https://github.com/devonfw/devon4j/issues/385

Acceptance-Criteria:
* [ ] class deleted https://github.com/devonfw/devon4j/blob/682c72e6440c9172806ae6393c553b2dc1853459/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/logic/impl/config/DefaultRolesPrefixPostProcessor.java#L16
* [ ] bean removed https://github.com/devonfw/devon4j/blob/682c72e6440c9172806ae6393c553b2dc1853459/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/service/impl/config/WebSecurityBeansConfig.java#L39
* [ ] `@RolesAllowed` annotation still working and spring-security is matching AccessControls from the permission annotated via `@RolesAllowed` without any artificial `ROLE_` prefix.

tobka777:
fixes #400

### DIFF
--- a/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
+++ b/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/common/impl/security/BaseUserDetailsService.java
@@ -51,7 +51,7 @@ public class BaseUserDetailsService implements UserDetailsService {
   }
 
   /**
-   * @param user the login of the user
+   * @param user the {@link UserDetails} from spring-security.
    * @return the associated {@link GrantedAuthority}s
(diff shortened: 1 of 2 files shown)


hohwille (approved):
@tobka777 thanks for this nice improvement. Great that we can even avoid the workaround of removing the "ROLE_" prefix. 👍

Issue #393, opened 2021-05-21
devon4j JpaInitializer documentation not matching releases


hohwille:
This JavaDoc says that you need to add a `spring-boot-starter-jpa`:

https://github.com/devonfw/devon4j/blob/master/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/JpaInitializer.java#L7

However, there is no such starter and has never been:
https://repo1.maven.org/maven2/com/devonfw/java/starters/

So either we need to create such starter or change the documentation and explaing how to manually configure the @Bean properly to make this work... Should also be referenced from [JpaHelper](https://github.com/devonfw/devon4j/blob/master/modules/jpa-basic/src/main/java/com/devonfw/module/jpa/dataaccess/api/JpaHelper.java) JavaDoc as otherwise this class is not usable and will  fail with errors like:
```
java.lang.IllegalStateException: EntityManager has not yet been initialized!
	at com.devonfw.module.jpa.dataaccess.api.JpaEntityManagerAccess.getEntityManager(JpaEntityManagerAccess.java:38)
	at com.devonfw.module.jpa.dataaccess.api.JpaHelper.asEntity(JpaHelper.java:49)
	at org.example.app.ordermanagement.dataaccess.api.ItemEntity.setOrderId(ItemEntity.java:102)
...
```

hohwille:
No starter required. All comes with the archetype:
* main: https://github.com/devonfw/devon4j/blob/master/templates/server/src/main/resources/archetype-resources/core/src/main/java/__packageInPathFormat__/general/dataaccess/impl/JpaConfig.java#L24
* test: https://github.com/devonfw/devon4j/blob/master/modules/test-jpa/src/main/java/com/devonfw/module/test/common/base/DbTest.java#L32

So we need to...
* update the documentation.
[...]

PR #403, opened 2021-06-18
#393: improved/fixed idref doc

hohwille:
Documentation for #393


hohwille:
For the record: This PR is only changing documentation. I will create another PR for changing the base test-classes as suggested in #393.

maihacke (approved):
Good job..

PR #410, opened 2021-07-06
#284: support for synchronous rest client using http client

hohwille:
Implementation of issue #284 (synchronous service client via HttpClient) and replacement for PR #401.
* refactored code to avoid C&P redundancies
* fixed @since tags to proper release `2021.04.003`
* small cleanup/improvement of existing service-client code

For the record: due to the refactoring, I changed some internal classes and methods. The public usage API remains unchanged but in case some project hooked in deeply to override code my changes could potentially break something. I doubt that anybody would do this from outside without getting in touch so IMHO all fine.


starters/pom.xml
```diff
@@ -28,6 +28,7 @@
     <module>starter-kafka-sender</module>
     <module>starter-kafka-receiver</module>
     <module>starter-http-client-rest-async</module>
+    <module>devon4j-starter-http-client-rest-sync</module>
```
hohwille, line 31: @SchettlerKoehler noticed that we are breaking our DRY conventions here.
We should therefore remove the `devon4j-` prefix from the folder name as we do for all other starters.

hohwille:
For the record: @ssarmokadam is testing this feature and will give feedback till end of this week so I plan to merge and release on monday.

hohwille:
@ssarmokadam found a stupid bug I made during the refactoring in her manual tests of the archetype that I fixed with https://github.com/devonfw/devon4j/pull/410/commits/93f6ff98531d60aa206639bb176f95e059353c5c

hohwille:
Also we wanted to follow opt-in approach for archetype rather than opt-out. Therefore in the core module of the archetype I...
* outcomented the dependency `devon4j-starter-cxf-server-ws` as providing web-services is rather outdated
* outcommented all `devon4j-starter-*-client-*` dependencies as service-client is only used by specific projects
* aligned the `devon4j-starter-*-client-*` dependencies with the new stuff and documentation

hohwille:
For the record:
@ssarmokadam made a manual test and found a bug using such REST operation:
```
@GET
  @Path("/clientrequest/")
  @Consumes(MediaType.APPLICATION_JSON)
  @Produces(MediaType.APPLICATION_JSON)
  public String returnResponseToClient(String args);
```

Call this worked via CXF client but failed with HttpClient with:
```
aused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (String)""; line: 1, column: 0]
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59) ~[jackson-databind-2.12.2.jar:2.12.2]
```

It turned out that the content-type was expected but missing.
I therefore fixed the MediaType/ContentType handling in my last commit: https://github.com/devonfw/devon4j/pull/410/commits/db6abd71287796103719a5bcbd50c07a6d534c5c

@ssarmokadam confirmed that my last commit fixes the problem.
Therefore I consider this PR as tested and ready to merge now.

Issue #420, opened 2021-07-19
modernization and alignment with devon4quarkus


hohwille:
We are in the process to offer two general options in the Java stack of devonfw:
* spring-boot
* quarkus

The current `devon4j` modules will remain the `spring-boot` specific libraries. We will fully support and maintain them also in the future.
In case we will create new modules/libraries for `quarkus` we will create those in https://github.com/devonfw/devon4quarkus

In the documentation we have two options:
1. Create and maintain a new and completely independet documentation for devon4quarkus and keep the `devon4j` documentation untouched as "devon4spring" documentation.
2. Adopt the existing `devon4j` documentation as a general documentation of the Java stack in devonfw. Separate options for spring/spring-boot and quarkus where things differ.

Both options have their pros and cons.
However, IMHO on the long run option 1. is better, because:
* It is reducing redundancies (~50% of the content with coding conventions, implementing a REST service with JAX-RS, etc. will be identical).
* Therefore maintenance effort is significantly reduced with 1.
* When searching for a specific topic with 1. you will typically get one hit for Java where as with 2. you will always get two hits even if both hits have the same content.

Please note that for the end-user we need a new and better way to approach the documentation anyway.
A user should choose for technology such as `C#` or `Java` on our devonfw.com website. When `Java` is selected we should give a brief introduction about the two options spring and quarkus with some pros and cons. From these two options you should be guided to the details via simple link to some nice summary guides explaining spring-boot and quarkus approach in Java stack of devonfw.

Therefore I create this issue to link first PRs in extending the `devon4j` documentation accordingly.

For the record: We can still have two separate PDFs generated for spring (classic devon4j) and quarkus (devon4quarkus) if we want (even when going for option 1.).

hohwille:
auto-closed by PR #421 but still not complete.

hohwille:
We also need to align the logging guide. There is no reason why we have an additional guide for quarkus that is 99% not quarkus specific:
* [devon4j logging guide](https://github.com/devonfw/devon4j/blob/master/documentation/guide-logging.asciidoc)
* [quarkus logging guide](https://github.com/devonfw/devon4j/blob/master/documentation/quarkus/guide-logging.asciidoc)

IMHO we need a facelift and alignment of the devon4j guide getting rid of the log files explanation moving to a cloud-native approach where apps just output JSON log to standard out what is already established in classic devon4j since > 1 year.
Then we can move the guide to general and delete the quarkus specific one when all is merged.

[...]

PR #430, opened 2021-08-05
add quarkus monitoring guide

lilliCao:
Relate #420 
Closes https://github.com/devonfw/devon4quarkus/issues/10

### DIFF
--- a/documentation/_Sidebar.asciidoc
+++ b/documentation/_Sidebar.asciidoc
@@ -64,3 +64,4 @@
 === Guides
 * link:quarkus/guide-logging.asciidoc[Logging]
 * link:quarkus/guide-cors-support.asciidoc[CORS support]
+* link:quarkus/guide-monitoring.asciidoc[Monitoring]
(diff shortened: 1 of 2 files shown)


GuentherJulian:
@hohwille has already generalized the original [monitoring guide](https://github.com/devonfw/devon4j/blob/master/documentation/guide-monitoring.asciidoc). There are no more spring specific things mentioned, just a link to the spring-boot-actuator guide. So I'm not sure if we still need that one or if we should keep the general guide with the link to the Quarkus guide. If we add the additional guide, we should also mention the smallrye-health extension, since health check is mentioned as one type of monitoring in the general guide.

lilliCao:
* @GuentherJulian yeah I see. I just don't think we should merge any quarkus-stuff there as I followed the 2. approach [here](https://github.com/devonfw/devon4j/issues/420). This guide is more or less based on an original tutorial from quarkus which I also linked to. But if we don't want to re-explain stuff, we can also simplify them and add a link to quarkus only.
* about smallrye-health: good point.

PR #431, opened 2021-08-06
Migration documentation added

suvmanda:
Migration documentation added


maybeec:
I don't agree to the "Nice to have" changes. And also there are things we need to discuss is, that this command just at the moment works if your project structure is kept as is `devon migrate` comes with a lot of restrictions.
I would even go further. Let's remove it here and provide a proper katacoda tutorial. I don't see the value as a plain text doc.

PR #432, opened 2021-08-06
Add guide for quarkus logic layer

lilliCao:
As discussed with @hohwille, we prefer to use the approach "use-cases without component-facade" for the logic layer implementation. This guide is based on the quarkus code example [here](https://github.com/devonfw-sample/devon4quarkus-reference/pull/7). 

With that said, this is a proposal and I would appreciate any feedback/discussion.

Relate https://github.com/devonfw-sample/devon4quarkus-reference/pull/7 https://github.com/devonfw/devon4j/issues/420
Closes https://github.com/devonfw-forge/devonfw-microservices/issues/38

### DIFF
--- a/documentation/_Sidebar.asciidoc
+++ b/documentation/_Sidebar.asciidoc
@@ -64,3 +64,4 @@
 === Guides
 * link:quarkus/guide-logging.asciidoc[Logging]
 * link:quarkus/guide-cors-support.asciidoc[CORS support]
+* link:quarkus/guide-logic-layer.asciidoc[Logic layer]
--- a/documentation/devon4j.asciidoc
+++ b/documentation/devon4j.asciidoc
@@ -274,3 +274,4 @@ include::guide-domain-layer.asciidoc[leveloffset=+2]
 === Guides
 
 include::quarkus/guide-cors-support.asciidoc[leveloffset=+3]
+include::quarkus/guide-logic-layer.asciidoc[leveloffset=+3]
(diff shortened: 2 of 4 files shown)


GuentherJulian:
Also add your guide to the [devon4j.asciidoc](https://github.com/devonfw/devon4j/blob/master/documentation/devon4j.asciidoc) file, so that the guide is displayed on the website

EduardKrieger (approved):

hohwille (changes requested):
@lilliCao thanks for your PR and taking this forward.
IMHO we still struggle to align on the strategy for the documentation.
Sorry, if I did not manage to explain it well enough.
I hope to be able to explain it better in our next standup.
See also [my review here](https://github.com/devonfw/devon4j/pull/435#pullrequestreview-725545768)

It should not depend on spring or quarkus how to implement a use-case. So we should keep this central guide and keep improving it:
https://github.com/devonfw/devon4j/blob/master/documentation/guide-usecase.asciidoc

What can be changed/improved:
* first of all the idea is that you create a class per use-case. For CRUD we wanted to simplify and inveted this `UcFind*` and `UcManage*`. However the focus of the guide should be on some real business logic so something like `UcApproveOrder` or `UcTerminateContract`.
* for simplicity we can state that creating interfaces for the use-cases is optional.
* Also I just re-read the existing usecase guide and was kind of shocked that the "Implementation" part is entirely broken and full of nonsense for 2 years+. We should take out the component-facade stuff that belongs [here](https://github.com/devonfw/devon4j/blob/master/documentation/guide-component-facade.asciidoc).

lilliCao:
Moved to #439

PR #433, opened 2021-08-06
Refactoring validation guide

GuentherJulian:
Closes https://github.com/devonfw/devon4quarkus/issues/12
Since there are not much differences between validation in Spring and in Quarkus, I have changed the original validation guide and moved it to the _General_ section of the website. Also changed the sidebar to match the structure of the website.


EduardKrieger (approved):

documentation/_Sidebar.asciidoc
```diff
@@ -1,8 +1,26 @@
...
+* link:guide-sql.asciidoc[SQL]
+* link:guide-database-migration.asciidoc[Database Migration]
+* link:guide-logging.asciidoc[Logging]
+* link:guide-monitoring.asciidoc[Monitoring]
+* link:guide-log-monitoring.asciidoc[Log-Monitoring]
+* link:guide-apm.asciidoc[Application Performance Management]
+* link:guide-security.asciidoc[Security]
 
-=== link:architecture.asciidoc[Architecture Overview]
+== spring
```
hohwille, line 23: indeed the sidebar is still messed, but also these changes do not completely fix this.
We should address this outside of this PR that is about validation guide...

hohwille (approved):
@GuentherJulian thanks for your changes 👍

documentation/_Sidebar.asciidoc
GuentherJulian: Removed the changes in the sidebar

documentation/guide-validation.asciidoc
```diff
@@ -30,10 +32,23 @@ A description of how to enable BV can be found in the relevant http://docs.sprin
...
+----
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-hibernate-validator</artifactId>
+    </dependency>
+----
+
+* For methods to validate go to their declaration and add constraint annotations to the method parameters.
++
+__In spring applications you can add the `+@Validated+` annotation to the implementation (spring bean) to be validated (this is an annotation of the spring framework, so it`s not available in the Quarkus context). The standard use case is to annotate the logic layer implementation, i.e. the use case implementation or component facade in case of simple logic layer pattern. Thus, the validation will be executed for service requests as well as batch processing.__
```
hohwille, line 46: FYI: asciidoc had an old syntax for code by sourrounding with plus symbols (e.g. +@Validated+).
The new syntax for code that goes inline with text is to sourround with back-ticks (e.g. `@Validated`).
You should remove the plus signs after chaning to backticks. We still have leftovers in our older doc sections using the plus syntax that should be replaced with backticks. Also something worth to share with the team in the standup. Also it would be great if you could find some minutes to add this here: https://github.com/devonfw/docgen/blob/master/documentation/guidelines.asciidoc

PR #439, opened 2021-08-12
adjust logic layer documentation

lilliCao:
* guide-logic-layer: rewrite the recommendation to "use-case approach"
* guide-component-facade": add note that "use-cases approach" is the recommended option
* guide-usecase: rewrite the implementation section and remove injection-issue section

Relate devonfw-sample/devon4quarkus-reference#7 #420


documentation/guide-usecase.asciidoc
```diff
@@ -53,23 +56,34 @@ public interface UcApproveBooking {
...
 [source,java]
 ----
-public interface Bookingmanagement extends UcFindBooking, UcManageBooking, UcApproveBooking {
+public class UcFindBookingImpl implements UcFindBooking {
+
+  @Inject
+  private BookingRepository bookingRepository;
+
+  @Override
+  public BookingEto findBooking(Long id) {
```
GuentherJulian, line 71: Didn't we want to change from the concept with `Eto` and `Cto` objects to a simple concept with `Dto`s in the future? Then we should adjust the code examples in this guide and maybe add a link to an example with the old concept, so readers who only know the old style won't be confused

GuentherJulian (changes requested):

documentation/guide-usecase.asciidoc
```diff
@@ -53,23 +56,34 @@ public interface UcApproveBooking {
 ----
 
 == Implementation
-For the implementation of a use-cas, the same rules that are described for the link:guide-component-facade.asciidoc#implementation[component-facade implementation].
-
- .
-
-However, when following the use-case approach, your component facade simply changes to:
+For the implementation of a use-case, the same rules that are described for the link:guide-component-facade.asciidoc#implementation[component-facade implementation].
```
hohwille, line 59: As we do not recommend this concept anymore, I would not like to it here. Maybe we can just drop this sentence as eveything follows in the next one.

documentation/guide-usecase.asciidoc
```diff
@@ -53,23 +56,34 @@ public interface UcApproveBooking {
...
 
+In short, the implementation should carry its own name and the suffix `Impl` and is annotated with `@Named`. It will need access to the persistent data which is done by injecting the corresponding repository (or DAO). Furthermore, it shall not expose persistent entities from the data access layer and has to map them to transfer objects using the bean-mapper. Please refer to our bean mapping documentation for more information.
```
hohwille, line 61: We also need `@ApplicationScoped`. However, I would link to `guide-dependency-injection` here and just have a fully functional sample code so add the `@ApplicationScoped` and `@Named` to the sample impl below.

documentation/guide-usecase.asciidoc
hohwille: That is a kind of tricky point.
To make our guide less biased, I would sugget to come up with a more general example that returns just any transfer-object (`*To`) and then for details refer to the `guide-transferobject.asciidoc`. This guide should again be copied to a `guide-eto-cto.asciidoc` and we can add a `guide-dto.asciidoc`. Then we can rewrite the `guide-transferobject.asciidoc` to only explain the general idea of transfer objects and then link to the two options `ETO/CTO` and `DTO` (please note that neither ETO is spring specific nor DTO is quarkus specific so do not put into these sub-folders).
As I am running out of time, could you file a story for this?

hohwille (approved):
@lilliCao thanks for this PR. You rescued this guide that was broken for long 👍 
I added some review comments for further improvements but to make progress I will already merge. So please create an issue for the remaining improvements that can be addressed by a new PR. Thanks!

PR #458, opened 2021-09-21
guide-for-testing

TugbaDalmaz:


documentation/quarkus/guide-for-testing.asciidoc
```diff
@@ -0,0 +1,59 @@
+In Quarkus there are two test dependencies, the JUnit5 and the rest-assured. When using the JUnit5 it must be noted that the Surfire Maven Plugin is set, as the default version does not support the JUnit5. The quarkus-junit5 is required for testing as this provides the @QuarkusTest annotations that control the testing framework. The rest-assured is not required but is used for convenient testing of HTTP endpoints. Quarkus also provides an integration that automatically sets the correct URL without configuration.
```
EduardKrieger, line 1: Surefire

documentation/quarkus/guide-for-testing.asciidoc
```diff
@@ -0,0 +1,59 @@
...
+@QuarkusTests must not be mixed with other tests in a single execution run, such as a single Maven Surfire Plugin execution. This is because the @QuarkusTest, when running the entire execution flow, starts a Quarkus server and prevents other tests from starting their own Quarkus server. To work around this limitation the JUnit5 tag io.quarkus.test.junit.QuarkusTest is used.
+
+Source: https://quarkus.io/guides/getting-started-testing
```
EduardKrieger, line 59: I think you can also show or mention the ability to use [continious testing](https://quarkus.io/guides/continuous-testing) with the developement mode in quarkus.

documentation/quarkus/guide-for-testing.asciidoc
```diff
@@ -0,0 +1,59 @@
...
+
+1. control of the test port
+Quarkus listens on port 8080 by default and uses port 8081 when running tests. This has the advantage that tests can be run in parallel with running applications. It is also possible to configure the test ports. Here, http.test-port for HTTP and quarkus.http.test-ssl-port for HTTPS must be set up in the properties of the application. If the value is set to 0, then the operating system used assigns a random port. 
+Furthermore, Quarkus provides a REST Assured integration without any additional configuration. This uses the updated default port from REST Assured before executing the test. When using REST Assured, the connection and response timouts are set to 30 seconds, these can be changed with the quarkus.http.test-timeout property.
```
EduardKrieger, line 6: timeouts

EduardKrieger (changes requested):

GuentherJulian:
You simply link the Quarkus guide in your guide, which is fine in many cases. But we also need to see if there are any differences with the existing Testing Guide from devon4j. For example, the existing guide for Spring applications says to use AssertJ as the assertion framework. The Quarkus guide depends on REST Assured, if I understand correctly. So the existing guide is not a general guide for Spring and Quarkus, if we say in another guide that the user must use a different framework in Quarkus. Either we need to remove those parts of the general guide and create different guides for Spring and Quarkus. Or we need to check if it is possible to use AssertJ in Quarkus as well. But then we can't just refer to the original Quarkus guide.
In any case we need to link your guide in the general guide so that user can easily jump to your guide when they read something about testing in the general guide.

maybeec:
closes  devonfw/devon4quarkus#11

EduardKrieger (approved):

maybeec (approved):

PR #466, opened 2021-10-04
spellcheck & sync-docs on change of docs 

sujith-mn:


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,18 @@
+name: spellcheck 
+on:
+  push:
+    paths:
+       - '**.asciidoc'
+       - '**.md'
```
hohwille, line 6: if we also support markdown we should IMHO also support `**.adoc` (see e.g. our README.adoc).

hohwille (approved):
@sujith-mn thanks for this PR and the actions 👍 
This is great. Do we have some repo where this is already tested and live?
Seems that you also optimized our actions so that regular build is not triggered by documentation only commits and also doc actions are only triggered by doc changes.
Should the spellcheck happen inside the doc-sync or why is it a separate action? In order to make it fail independent of a successful doc-sync - seems so and makes sense.

I am happy to approve. However, if we support markdown we should also suppoport adoc extension. Did you also adapt the action templates in .github repo?

maybeec (approved):

SchettlerKoehler (approved):

Issue #480, opened 2021-10-19
spring to quarkus guide


SchettlerKoehler:
As a developer that has to convert a spring to a quarkus application, I want a guide on how to do that so that I can make use of the experiences of others.

PR #487, opened 2021-10-27
Guide about migration from Spring to Quarkus

GuentherJulian:
Closes https://github.com/devonfw/devon4j/issues/480
* Added guide for migration
* Added an own file for the performace comparision and linked it from the [getting-started-for-spring-developers guide](https://github.com/GuentherJulian/devon4j/blob/migration_guide_spring_quarkus/documentation/quarkus/getting-started-for-spring-developers.asciidoc) and the pros section of Quarkus


maybeec (commented):
Why don't we create this guide as a katacoda tutorial to be automatically generated into a step by step wiki tutorial as well? You started properly with starting with jtq for example, so you could provide a proper click-adventure easily or is there something I overlooked?

GuentherJulian:
> Why don't we create this guide as a katacoda tutorial to be automatically generated into a step by step wiki tutorial as well? You started properly with starting with jtq for example, so you could provide a proper click-adventure easily or is there something I overlooked?

I think migrating the whole jtq example would result in tutorial that is too big, since the application provides multiple REST APIs and use cases. But of course it would be very helpful. Maybe a simpler application or just one REST service from jtq.

maybeec (approved):

PR #530, opened 2022-01-05
spell check code re-usable

prathibhapadma:


sujith-mn:
spellcheck removed from all individual repository and kept a common spellcheck file at .github to reuse.

Same changes done for sync-docs too, a common spell check file added at .github which will be reused in all repos.

hohwille (approved):

PR #550, opened 2022-04-22
#551: spring4shell - spring upgrade

sujith-mn:


hohwille (approved):
@sujith-mn Excellent. Thanks for the PR with security fix.
For new projects "latest is greatest" strategy is perfect.
Existing projects could consider if smaller updates will do as well.
But for IDE migration we will then also upgrade to exactly these versions...

Issue #568, opened 2022-07-21
Cobigen generates ToTest which is not supporting SortedSet


maybeec:
@rhobolic commented on [Thu Jun 02 2022](https://github.com/devonfw/cobigen/issues/1540)

**Problem**
When we use SortedSet in the To the ToTest is failing in the method "private void testEqualsAndHashcode(Class<?> clazz, SoftAssertions assertion, ToExclusion exclusion)"

**Expectation**
The test should not fail for SortedSet

**Solution hint**
If the green undelayed snippet is added everything works fine again
![image](https://user-images.githubusercontent.com/99953472/171579166-035cb39a-8d39-48e4-b171-ac8fc3311295.png)

PR #570, opened 2022-07-27
Changed toTest to support SortedSet

NelsonNew:
#568


hohwille (approved):
@NelsonNew thanks for this PR. 👍 Change makes perfectly sense.

PR #573, opened 2022-08-26
Expanding Orika Documentation 

ThomasRot:
As we are currently recommending Orika and we supply `com.devonfw.module.beanmapping.common.impl.orika.CustomMapperEto` when using the default `getOrika`-Bean, I would recommend documenting, that:

- we already provide the mapping of `persistentEntity`, that is explicitly described in the Dozer-Config 
- proper further configuration can be supplied by extending `BaseOrikaConfig#configureCustomMapping` and there is no need to override `getOrika`


hohwille (approved):
@ThomasRot thanks for the PR. 👍 
Sorry, for the incredibly long delay. I was drowning in a critical project and nobody else to care of your nice contribution.
Ready for merge now.

PR #578, opened 2022-11-21
Feature/new page announcement

denise-khuu:


SchettlerKoehler (approved):

hohwille:
I disagree with this approach and will revert the changes.
We are currently creating maintenance releases for projects relying on devon4j.
The message should also be slightly different.

Issue #585, opened 2023-01-19
Conisder new CVE updates (cxf, jackson, snakeyaml, etc.)


hohwille:
In devon4j we need to do another update to close most recent CVEs:
* snakeyaml: 1.30 has several CVEs and an update is required (1.33). However, no spring-boot 2.x release addresses this but (so far) only 3.0.1 - see also #582
UPDATE: Even after updating snakeyaml to the currently latest version 1.33 it still has a high vulnerability left: CVE-2022-1471 So nothing currently possible to close this one. Can only be addressed, once another version of snakeyaml comes out fixing it.
* apache CXF has seen another set of CVEs that dont seem so critical but have been [hyped by BSI](https://www.news.de/technik/856646574/apache-cxf-gefaehrdet-it-sicherheitswarnung-vom-bsi-und-bug-report-betroffene-systeme-und-produkte-neue-versionen-und-updates/1/). Here an update to 3.5.5 (or even 4.0.0) would resolve the CVEs.
* update jackson to 2.14.1
* ~also junit should be updated (CVE-2022-31514)~

hohwille:
Another thing to consider is that our BOM imports the BOM of spring-cloud.
However, IMHO spring-cloud-dependencies should be questioned from security PoV:
https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-dependencies/

We introduced this for kafka. Our kafka module is more or less deprecated. Hence, we should consider getting rid of this large dependency-tree that can cause more harm than use.

PR #587, opened 2023-01-19
#585: get rid of spring-cloud, disable kafka module, updates 

hohwille:
addresses #585 with:
* kick out kafka module (no more releases of devon4j-kafka, please use spring kafka module instead)
* remove spring-cloud from our dependency-management and BOM
* update spring-boot to 2.7.7
* update jackson to 2.14.1


hohwille:
Wow, was merged after successful build, but actually build only succeeded since SNAPSHOT versions had been build in a previous release and retrieved from OSSRH nexus. I fixed this in PR #588 as this was already merged.

